PR #5, opened 2016-07-04
Added an overload of File() that accepts an ITextFormatter

### LINKED ISSUE #4: How to write JSON to file?
Is there a way to write the structured log to file (in json)?

nblumhardt:
Fixes #4. Usage is:

```csharp
.WriteTo.File(new JsonFormatter(), "C:\\path", ...<as before>)
```

The existing signature accepts both an output template and format provider, to create `MessageTemplateTextFormatter` under the hood. This one omits both of those parameters and accepts the formatter directly as the first arg.

There are two options for positioning: first, before the path, or second, after it. Not sure if I have a strong preference either way - adding it first gets rid of the intellisense ambiguity as soon as the user has specified a path, so might be the cleaner option in usage.

Once merged this should publish a `2.1.0-dev-nnn` package version.


nblumhardt:
Ready to merge, just in case the last sentence of the PR description created any doubt :-)

adamchester:
:boom:

PR #6, opened 2016-07-04
2.1 Release

nblumhardt:
* Adds an overload to `WriteTo.File()` that accepts an `ITextFormatter` (#4)

Using this as an opportunity to test out the build/release scripts under dev/master scenarios. If all goes well (apart from the issue of tag names, TBD), these should form the basis for some shared build infrastructure.


nblumhardt:
Kicking this one off, lots to do today.

Issue #9, opened 2016-07-21
Default UTF-8 encoding should not include BOM


nblumhardt:
https://github.com/serilog/serilog-sinks-file/blob/dev/src/Serilog.Sinks.File/Sinks/File/FileSink.cs#L59

The built-in `Encoding.UTF8` object is problematic because it emits the byte order mark not otherwise required or expected in UTF-8 encoded files.

This should instead be a:

```csharp
static readonly Encoding DefaultUtf8 = new UTF8Encoding(false);
```

PR #10, opened 2016-07-26
Changed the UTF Encoding instance to avoid inserting the BOM character

christianhxc:
Fix to solve issue #9

### DIFF
--- a/src/Serilog.Sinks.File/Sinks/File/FileSink.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/FileSink.cs
@@ -56,7 +56,7 @@ namespace Serilog.Sinks.File
             TryCreateDirectory(path);
 
             var file = System.IO.File.Open(path, FileMode.Append, FileAccess.Write, FileShare.Read);
-            var outputWriter = new StreamWriter(file, encoding ?? Encoding.UTF8);
+            var outputWriter = new StreamWriter(file, encoding ?? new UTF8Encoding(false));
             if (fileSizeLimitBytes != null)
             {
                 var initialBytes = file.Length;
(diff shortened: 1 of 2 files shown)


christianhxc:
@nblumhardt it seems that this won't be possible for now, look https://github.com/dotnet/coreclr/issues/5000

nblumhardt:
Thanks Christian!

I don't think that the linked issue is the problem here, the compiler error is:

```
C:\projects\serilog-sinks-file\src\Serilog.Sinks.File\Sinks\File\FileSink.cs(59,71): error CS0246: The type or namespace name 'UTF8Encoding' could not be found (are you missing a using directive or an assembly reference?)
```

If the constructor accepting `bool` isn't present, the issue suggests that `new UTF8Encoding()` will do as we desire - a conditional compilation directive would cover both platforms in that case.

christianhxc:
@nblumhardt still not working, that constructor doesn't exist anymore

nblumhardt:
The package:

```json
        "System.Text.Encoding.Extensions": "4.0.11"
```

needs to be added to the netstandard1.3 dependency group for the `UTF8Encoding` type to appear.

Just a thought, shall we write it as:

```csharp
new UTF8Encoding(encoderShouldEmitUTF8Identifier: false)
```

I.e., name the parameter so that there's absolutely no ambiguity?

The above works without any conditional compilation, so seems like the way to go :+1:

christianhxc:
@nblumhardt not sure what happened but yesterday I just tried the same thing and it didn't work (guess it was my workstation). Anyways, I already submitted the change here, I'm gonna do it for the RollingFile project :)

nblumhardt:
Thanks @christianhxc, great stuff :-)

PR #12, opened 2016-08-16
Enable auditing

nblumhardt:
serilog/serilog#828

This makes `AuditTo.File()` work with Serilog 2.2-*.

### DIFF
--- a/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
+++ b/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
@@ -95,26 +95,96 @@ namespace Serilog
             LoggingLevelSwitch levelSwitch = null,
             bool buffered = false)
         {
+            return ConfigureFile(sinkConfiguration.Sink, formatter, path, restrictedToMinimumLevel, fileSizeLimitBytes, levelSwitch, buffered);
+        }
+
+        /// <summary>
+        /// Write log events to the specified file.
+        /// </summary>
+        /// <param name="sinkConfiguration">Logger sink configuration.</param>
+        /// <param name="path">Path to the file.</param>
+        /// <param name="restrictedToMinimumLevel">The minimum level for
+        /// events passed through the sink. Ignored when <paramref name="levelSwitch"/> is specified.</param>
+        /// <param name="levelSwitch">A switch allowing the pass-through minimum level
+        /// to be changed at runtime.</param>
+        /// <param name="formatProvider">Supplies culture-specific formatting information, or null.</param>
+        /// <param name="outputTemplate">A message template describing the format used to write to the sink.
+        /// the default is "{Timestamp} [{Level}] {Message}{NewLine}{Exception}".</param>
+        /// <returns>Configuration object allowing method chaining.</returns>
+        /// <remarks>The file will be written using the UTF-8 character set.</remarks>
+        public static LoggerConfiguration File(
+            this LoggerAuditSinkConfiguration sinkConfiguration,
+            string path,
+            LogEventLevel restrictedToMinimumLevel = LevelAlias.Minimum,
+            string outputTemplate = DefaultOutputTemplate,
+            IFormatProvider formatProvider = null,
+            LoggingLevelSwitch levelSwitch = null)
+        {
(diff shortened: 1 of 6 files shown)


src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
```diff
@@ -95,26 +95,96 @@ public static class FileLoggerConfigurationExtensions
...
             if (formatter == null) throw new ArgumentNullException(nameof(formatter));
             if (path == null) throw new ArgumentNullException(nameof(path));
+            if (fileSizeLimitBytes.HasValue && fileSizeLimitBytes < 0) throw new ArgumentException("Negative value provided; file size limit must be non-negative");
 
             FileSink sink;
             try
             {
                 sink = new FileSink(path, formatter, fileSizeLimitBytes, buffered: buffered);
             }
-            catch (ArgumentException)
```
nblumhardt: The precondition checks have always thrown in `File()`; this PR changes what throws/what doesn't very slightly, in that invalid path errors are now always suppressed in `WriteTo` scenarios and always propagate with `AuditTo`.

In general it's really hard to pin down which argument errors should be throwing ones. Serilog aims to always propagate API usage errors, and always (by default) suppress environmental/transient/runtime errors. In the case of the System.IO APIs and various configuration providers the line is fuzzy at times.

src/Serilog.Sinks.File/Sinks/File/FileSink.cs
```diff
@@ -53,7 +52,11 @@ public FileSink(string path, ITextFormatter textFormatter, long? fileSizeLimitBy
             _textFormatter = textFormatter;
             _buffered = buffered;
 
-            TryCreateDirectory(path);
+            var directory = Path.GetDirectoryName(path);
+            if (!string.IsNullOrWhiteSpace(directory) && !Directory.Exists(directory))
+            {
+                Directory.CreateDirectory(directory);
```
nblumhardt: These errors were previously handled locally; now because there is more suppression in the caller we don't need the additional EH here.

Note that this change doesn't modify the overall behaviour of this constructor, because previously an initial exception, though handled and logged, would always be immediately followed with another exception from `Open` on the line following.

test/Serilog.Sinks.File.Tests/FileLoggerConfigurationExtensionsTests.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
+using Serilog;
+using Serilog.Sinks.File.Tests.Support;
+using Serilog.Tests.Support;
+using Xunit;
+
+namespace Serilog.Tests
+{
+    public class FileLoggerConfigurationExtensionsTests
```
nblumhardt: File IO failures are hard to emulate in tests - invalid paths and broken formatting should be close enough to IO errors to give a reasonable measure of confidence.

src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
```diff
@@ -95,26 +95,96 @@ public static class FileLoggerConfigurationExtensions
...
+        /// <returns>Configuration object allowing method chaining.</returns>
+        /// <remarks>The file will be written using the UTF-8 character set.</remarks>
+        public static LoggerConfiguration File(
+            this LoggerAuditSinkConfiguration sinkConfiguration,
+            ITextFormatter formatter,
+            string path,
+            LogEventLevel restrictedToMinimumLevel = LevelAlias.Minimum,
+            LoggingLevelSwitch levelSwitch = null)
+        {
+            return ConfigureFile(sinkConfiguration.Sink, formatter, path, restrictedToMinimumLevel, null, levelSwitch, false, true);
```
nblumhardt: Note here that auditing 1) does not allow buffered writes, and 2) doesn't support file size limiting. Seems like the safest starting point.

We could probably enable file size limits and just throw when they're hit, but I'm okay with leaving this alone for the first cut.

nblumhardt:
Merging so that the downstream items (XML and JSON support) can be kicked off :-)

PR #16, opened 2016-08-30
Make shared log files readable while in use

nblumhardt:
Missed the `FileShare.Read` option in the original implementation :-O

Fixes https://github.com/serilog/serilog-sinks-rollingfile/issues/20

### DIFF
--- a/src/Serilog.Sinks.File/Sinks/File/FileSink.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/FileSink.cs
@@ -14,9 +14,6 @@
 
 using System;
 using System.IO;
-#if ATOMIC_APPEND
-using System.Security.AccessControl;
-#endif
 using System.Text;
 using Serilog.Core;
 using Serilog.Events;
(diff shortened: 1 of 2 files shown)


src/Serilog.Sinks.File/Sinks/File/FileSink.cs
```diff
@@ -64,13 +61,7 @@ public FileSink(string path, ITextFormatter textFormatter, long? fileSizeLimitBy
                 Directory.CreateDirectory(directory);
             }
 
-#if ATOMIC_APPEND
-            // FileSystemRights.AppendData improves performance substantially (~30%) when available.
```
nblumhardt: Made an error while benchmarking: the improvement was due to another change brought in accidentally (the removal of `CharacterLimitingTextWriter`, not this flag, so reverting the little accommodation here.

nblumhardt:
Pushing through to dev so that we can get some wider usage, it'd be nice to send this on towards master as soon as possible :-)

PR #20, opened 2016-10-05
Flush to disk option

### LINKED ISSUE #18: Flush-to-disk option
In Windows 10 and Windows Server 2012 R2, StreamWriter.Flush won't write the data to disk immediately.

You can know this by running under code and watching the file size through "dir" command.
```
Log.Logger = new LoggerConfiguration()
    .WriteTo.File("log.txt")
    .CreateLogger();

for (var i = 0; i < 100; ++i)
{
    Console.WriteLine(i);
    Log.Information("Hello, file logger!");
    System.Threading.Thread.Sleep(1000);
}
```

We can write to disk immediately by touching the file like this.
```
var touched = new FileInfo("log.txt").LastWriteTime;
```

Could you please improve "buffered:false" or add new option to do this?

P.S.
I'm using Serilog.Sinks.RollingFile for windows service logging, and collecting the logs by Microsoft Operations Management Suite (OMS). I found OMS log collection is not working good, because above.

nblumhardt:
Fixes #18 (rolling file support needed downstream).

```csharp
Log.Logger = new LoggerConfiguration()
      .WriteTo.File("log.txt", flushToDiskInterval: TimeSpan.FromSeconds(1))
      .CreateLogger();
```

Attempts to optimize things a bit so no flush is performed unless events have been written.

The `IFlushableFileSink` and `PeriodicFlushToDiskSink<T>` are (reluctantly) public, as this will permit the downstream `RollingFileSink` to wire up to the same mechanism, without us having to encapsulate the periodic flushing behavior in `FileSink` and `SharedFileSink` directly. In the long run, we might get some kind of generialized `IFlushable` into the base Serilog package and make it available across the board through some kind of generic wrapper.

I verified by hand that the `FlushToDisk()` is called, but can't come up with an easy way to test it without some kind of multi-process runner. If we query the file size from Windows from the unit test process, it'll always report a consistent view of the world to us :-).

**Bonus:** discovered the shared file sink tests were inadvertently ignored 😢; after enabling, found a bug preventing size limiting from taking effect. Fixed this too.

### DIFF
--- a/src/Serilog.Sinks.File/Sinks/File/PeriodicFlushToDiskSink`1.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/PeriodicFlushToDiskSink`1.cs
@@ -1,0 +1,66 @@
+﻿using System;
+using System.Threading;
+using Serilog.Core;
+using Serilog.Debugging;
+using Serilog.Events;
+
+namespace Serilog.Sinks.File
+{
+    /// <summary>
+    /// A sink wrapper that periodically flushes the wrapped sink to disk.
+    /// </summary>
+    /// <typeparam name="TSink">The type of the wrapped sink.</typeparam>
+    public class PeriodicFlushToDiskSink<TSink> : ILogEventSink, IDisposable
+        where TSink : ILogEventSink, IFlushableFileSink
+    {
+        readonly TSink _sink;
+        readonly Timer _timer;
+        int _flushRequired;
+
+        /// <summary>
+        /// Construct a <see cref="PeriodicFlushToDiskSink{TSink}"/> that wraps
+        /// <paramref name="sink"/> and flushes it at the specified <paramref name="flushInterval"/>.
+        /// </summary>
+        /// <param name="sink">The sink to wrap.</param>
+        /// <param name="flushInterval">The interval at which to flush the underlying sink.</param>
+        /// <exception cref="ArgumentNullException"></exception>
+        public PeriodicFlushToDiskSink(TSink sink, TimeSpan flushInterval)
+        {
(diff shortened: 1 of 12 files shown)


nblumhardt:
CC @pakrym

src/Serilog.Sinks.File/Sinks/File/PeriodicFlushToDiskSink`1.cs
```diff
@@ -0,0 +1,66 @@
...
+using Serilog.Debugging;
+using Serilog.Events;
+
+namespace Serilog.Sinks.File
+{
+    /// <summary>
+    /// A sink wrapper that periodically flushes the wrapped sink to disk.
+    /// </summary>
+    /// <typeparam name="TSink">The type of the wrapped sink.</typeparam>
+    public class PeriodicFlushToDiskSink<TSink> : ILogEventSink, IDisposable
```
pakrym: Why generic?

pakrym:
LGTM

src/Serilog.Sinks.File/Sinks/File/PeriodicFlushToDiskSink`1.cs
nblumhardt: Yeah, I'm not so keen on this upon second thoughts. It is to make clear that the passed-in `ILogEventSink` should also be `IFlushableFileSink` (no union types in C#) but the code would be a bit nicer to read if non-generic and tested at runtime. Going to change this.

nblumhardt:
Okay, I'm going to make a couple of minor tweaks this morning, then push through a `-dev` version and follow up with the `RollingFile` version. Should be all through as pre-release later today.

Issue #23, opened 2016-11-11
Extra BOM bytes being inserted in Seq buffer file


EamonHetherton:
This has taken some time to investigate, and seems that the problem is due to a confluence of issues across separate Seq/serilog libraries 

The problem description:
We are using serilog to log to Seq using the DurableSeqSink with a file size limit set.  For a particular use case we have implemented an web endpoint that has the same api as Seq raw events endpoint. 

What we were finding is that the data being posted often had an extra "}," (close brace and comma) like this:
{
  "Events": [{
    "Timestamp": "2015-05-09T22:09:08.12345+10:00",
    "Level": "Warning",
    "MessageTemplate": "Disk space is low on {Drive}",
    "Properties": {
      "Drive": "C:",
      "MachineName": "nblumhardt-rmbp"
    }
  },
  }
]
}
which meant that the payload was not valid json and so would fail to deserialize and return an error.

Eventually I determined that the buffer file for the serilog seq logger had additional bytes (invisible in a normal text editior).  These were the bytes "ef bb bf" which happens to be the UTF8 BOM.  

Again, while not ideal, this would not have caused the problem but for the fact that the HttpLogShipper does it's own byte counting instead of just relying on StreamReader (which seems to handle the extra BOM part way through a file) and it assumes that the BOM can only exist at the start of the file so it ends up incorrectly positioning its cursor 3 bytes before the end of the line instead of at the beginning of the next line, hence the extra '}'.

Now, as to why the extra bytes get into the file in the first place, it turns out that due to a quirk in the StreamWriter implementation ([https://referencesource.microsoft.com/#mscorlib/system/io/streamwriter.cs,208](url)) that only skips writing the encoding preamble when appending to a stream if the stream is seekable, when you set a fileSizeLimitBytes on the FileSink the preamble gets written again.

When using the DurableSeqSink, it uses one file per day by default so we get the error as manytimes as the process starts on a given day.

I have created some unit tests to demonstrate the problem (pull request coming) and I can think of several ways to work around the problem, but ultimately I think that coordinated changes need to be made to both the Serilog.Sinks.File.FileSink and Serilog.Sinks.Seq.HttpLogShipper.

[...]

PR #24, opened 2016-11-11
Issue #23 workaround

EamonHetherton:
Tests to demonstrate the issue and a workaround for it.  I wouldn't consider the workaround a final solution though.

### DIFF
--- a/src/Serilog.Sinks.File/Sinks/File/WriteCountingStream.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/WriteCountingStream.cs
@@ -47,7 +47,7 @@ namespace Serilog.Sinks.File
 
         public override void Flush() => _stream.Flush();
         public override bool CanRead => false;
-        public override bool CanSeek => false;
+        public override bool CanSeek => _stream.CanSeek;
         public override bool CanWrite => true;
         public override long Length => _stream.Length;
 
@@ -60,7 +60,7 @@ namespace Serilog.Sinks.File
 
         public override long Seek(long offset, SeekOrigin origin)
         {
-            throw new NotSupportedException();
+            return _stream.Seek(offset, origin);
         }
 
         public override void SetLength(long value)
--- a/test/Serilog.Sinks.File.Tests/FixedOutputFormatter.cs
+++ b/test/Serilog.Sinks.File.Tests/FixedOutputFormatter.cs
@@ -1,0 +1,25 @@
+﻿using Serilog.Formatting;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Serilog.Events;
+using System.IO;
+
+namespace Serilog.Tests
+{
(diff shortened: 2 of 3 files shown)


src/Serilog.Sinks.File/Sinks/File/WriteCountingStream.cs
```diff
@@ -60,7 +60,7 @@ public override long Position
 
         public override long Seek(long offset, SeekOrigin origin)
         {
-            throw new NotSupportedException();
+            return _stream.Seek(offset, origin);
```
nblumhardt: Would the workaround be impacted if we continue to throw here? Although it appears to break the `CanSeek`/`Seek()` contract, actually calling `Seek()` would break the write counting. Perhaps as a compromise we could throw `InvalidOperationException()` instead?

test/Serilog.Sinks.File.Tests/FixedOutputFormatter.cs
```diff
@@ -0,0 +1,25 @@
+﻿using Serilog.Formatting;
```
nblumhardt: Convention in the test project is for these to go into the `Support` namespace/folder.

nblumhardt:
Hi Eamon, I will make these changes by hand on the way "through the works" - thanks again for the investigation and PR.

EamonHetherton:
Thanks Nic, I unplugged for the weekend so would have got to this today otherwise.

nblumhardt:
Unplugging over the weekend is a good policy, I try to do the same :-) ... If you have a chance to spin up the 3.1.1 dev build it would be great to get the extra validation that all is working smoothly now. Cheers!

EamonHetherton:
just a few quick checks with Serilog.Sinks.File 3.1.1-dev-00754 and the issue appears solved with this version.

nblumhardt:
Great, thanks. #25 will push this through to master once I've done some more testing on different configurations here.

src/Serilog.Sinks.File/Sinks/File/WriteCountingStream.cs
EamonHetherton: Is the WriteCountingStream actually necessary at all?  For what I can see, being a private sealed class with only a single usage within the project, it exists only to support a file size limit.  In this case it seems to me that a regular FIleStream would be sufficient by using the Length property and more robust to boot.  No functionality of the actual stream would need to be wrapped and turned of either, like Seek.  Am I missing something?

nblumhardt: It's a good question. `FileStream.Length` turns out to be a couple of orders of magnitude slower, as it calls through to get the file size from the OS (at least, that's my recollection from testing it).

Open to scrapping it, but would need to re-do benchmarking. There's a small sample app in the solution that gives some basic write throughput numbers if you're keen to check it out - I'll have another look when I get a chance 👍

PR #27, opened 2016-11-24
Multiprocess sharing on .NET Core

nblumhardt:
~~Adds a new configuration for `netcoreapp1.0`, as although `Mutex` is available on `netstandard1.3` I think it's unlikely to work on all platforms targeted by that TFM (UWP doesn't like _System.Threading_ references, for instance).~~

Performance is about 10x slower than atomic append, but otherwise appears to work as far as my limited tests have covered.

Required for serilog/serilog-sinks-rollingfile#37.

### DIFF
--- a/example/Sample/Program.cs
+++ b/example/Sample/Program.cs
@@ -14,7 +14,7 @@ namespace Sample
             var sw = System.Diagnostics.Stopwatch.StartNew();
 
             Log.Logger = new LoggerConfiguration()
-                .WriteTo.File("log.txt", shared: true)
+                .WriteTo.File("log.txt")
                 .CreateLogger();
 
             for (var i = 0; i < 1000000; ++i)
--- a/example/Sample/Program.cs
+++ b/example/Sample/Program.cs
@@ -11,11 +11,11 @@ namespace Sample
         {
             SelfLog.Enable(Console.Out);
 
-            Log.Logger = new LoggerConfiguration()
-                .WriteTo.File("log.txt")
-                .CreateLogger();
-
             var sw = System.Diagnostics.Stopwatch.StartNew();
+
+            Log.Logger = new LoggerConfiguration()
+                .WriteTo.File("log.txt", shared: true)
+                .CreateLogger();
 
             for (var i = 0; i < 1000000; ++i)
             {
--- a/src/Serilog.Sinks.File/Sinks/File/SharedFileSink.OSMutex.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/SharedFileSink.OSMutex.cs
@@ -66,8 +66,8 @@ namespace Serilog.Sinks.File
                 Directory.CreateDirectory(directory);
             }
 
-            // Backslash is special on Windows
-            _mutex = new Mutex(false, path.Replace('\\', ':') + MutexNameSuffix);
+            var mutexName = Path.GetFullPath(path).Replace(Path.DirectorySeparatorChar, ':') + MutexNameSuffix;
+            _mutex = new Mutex(false, mutexName);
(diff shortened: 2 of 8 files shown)


example/Sample/Program.cs
```diff
@@ -11,12 +11,12 @@ public static void Main(string[] args)
         {
             SelfLog.Enable(Console.Out);
 
+            var sw = System.Diagnostics.Stopwatch.StartNew();
+
             Log.Logger = new LoggerConfiguration()
-                .WriteTo.File("log.txt")
+                .WriteTo.File("log.txt", shared: true)
```
nblumhardt: I'll remove this from the example, since it's possible people will copy it into their projects and sacrifice performance for no reason.

nblumhardt:
Ignore the comment about UWP: it's _System.Threading.Thread_ it rejects. Will update the PR.

nblumhardt:
Alrighty, I think this is ready to go.

nblumhardt:
(Does also need a test on Linux. Assumed Travis was churning away as it does for serilog/serilog but not so. Might do that from dev along with macOS.)

src/Serilog.Sinks.File/Sinks/File/SharedFileSink.OSMutex.cs
```diff
@@ -0,0 +1,145 @@
...
+            if (logEvent == null) throw new ArgumentNullException(nameof(logEvent));
+
+            lock (_syncRoot)
+            {
+                if (!_mutex.WaitOne(MutexWaitTimeout))
+                {
+                    SelfLog.WriteLine("Shared file mutex could not be acquired in {0} ms for event emitting", MutexWaitTimeout);
+                    return;
+                }
+
```
skomis-mm: I believe you should also handle `AbandonedMutexException` in case of another process crash and also protect against async exceptions with help of return value of `WaitOne()`:
```csharp
bool ownsMutex = false;
try
{
    try
    {
        ownsMutex = _syncRoot.WaitOne(MutexWaitTimeout);
    }
    catch (System.Threading.AbandonedMutexException)
    {
        ownsMutex = true;
    }

    // ...
}
finally
 {
      if (ownsMutex)
      {
          _syncRoot.ReleaseMutex();
      }
 }
```

src/Serilog.Sinks.File/Sinks/File/SharedFileSink.OSMutex.cs
```diff
@@ -0,0 +1,145 @@
...
+            _fileSizeLimitBytes = fileSizeLimitBytes;
+
+            var directory = Path.GetDirectoryName(path);
+            if (!string.IsNullOrWhiteSpace(directory) && !Directory.Exists(directory))
+            {
+                Directory.CreateDirectory(directory);
+            }
+
+            // Backslash is special on Windows
+            _mutex = new Mutex(false, path.Replace('\\', ':') + MutexNameSuffix);
```
nblumhardt: Good point, will do.

src/Serilog.Sinks.File/Sinks/File/SharedFileSink.OSMutex.cs
nblumhardt: Thanks, I'll look into it 👍

nblumhardt:
Sorry about that, I should have made a note here - I just picked it up this afternoon when I realised how much file-related work was queued up :-)

My vague "10x" was with two processes writing on Windows. When there's contention it _really_ seems to slow down. Probably not noticeable in practice though.

Thanks for checking the code, I'll run back through it in the morning.

nblumhardt:
I think everything should be covered now.

skomis-mm:
Looks good 👍

PR #35, opened 2017-10-13
Roll on year/month/day/hour/minute, on file size, or on a combination of both

nblumhardt:
Fixes serilog/serilog-sinks-rollingfile#33, Fixes serilog/serilog-sinks-rollingfile#52.

This PR introduces some additional parameters to `WriteTo.File()`.

First, `rollOnFileSizeLimit`, causes additional log files to be created when the file size limit is reached. E.g.:

```csharp
    .WriteTo.File("log.txt", rollOnFileSizeLimit: true)
```

will roll the file from `log.txt` to `log_001.txt`, `log_002.txt` and so-on when `fileSizeLimitBytes` is reached (the default is 1 GB). Old files will be cleaned up as per `retainedFileCountLimit` - the default is 31.

A second parameter, `rollingInterval`, rolls the file by `RollingInterval.Year`, `Month`, `Day`, `Hour` or `Minute` into `log20180101.txt`, `log20180102.txt` and so on.

```csharp
    .WriteTo.File("log.txt", rollingInterval: RollingInterval.Hour)
```

If _neither_ parameter is supplied, the behavior is exactly as before - the file name is static and no additional files will ever be created.

If _both_ parameters are specified, the first file for a rolling interval will be named like `log20180101.txt`, rolling into `log20180101_001.txt`, `log20180101.txt_002.txt` and so on, until the next interval when the counter resets.

If _either_ parameter is supplied, and the log file can't be opened, the suffix `_NNN` will be appended to find a filename that is available.

### Goals

 * Address the need for roll-on-size
 * Bring the code for _RollingFileSink_ into the same package as _FileSink_ so that they can be extended and refactored in future (e.g. for many of the other long-standing items in https://github.com/serilog/serilog-sinks-rollingfile/issues)
 * Simplify the user experience (most users find _File_ first, but really want _RollingFile_ - very few apps use _File_
 * Move away from `-{Date}` path templates - they're misleading, as formatting information is not supported, and the placeholder can only appear in a very limited portion of the path
 * Re-use the rolling file sink code to get things unblocked quickly; the port to the new rolling scheme was very direct - where given the choice, I made the existing patterns work instead of introducing new code
 * Update to VS2017 tooling

### Non-goals

 * Introduce much more flexibility - the combined codebases need some attention before adding code
 * Replace _Serilog.Sinks.RollingFile_ - new apps should move to _File_, but there's no reason existing apps should move if the functionality over there is satisfactory

### Breaking changes/evolution

 * The `*Sink` classes added in this PR are internal; the existing ones like `FileSink` are now marked `[Obsolete]` so that they too can become internal in a far-future release, to support some refactoring and improvement of the architecture
 * The default output template has been modernized with `:u3` short level names and `:lj` message formatting (this is the reason for the major version bump)

I'm sure I'll think of more :-)


nblumhardt:
(Could use more test coverage; the tests are mostly for the scenarios that overlap with _Serilog.Sinks.RollingFile_ - will loop back and try to add some more when possible.)

optical:
Awesome work @nblumhardt! This will let me retire my forked sink with a messy implementation of these features.

ritasker:
Looks great thanks. 👍

nblumhardt:
Great, thanks for taking a look!

Any thoughts on naming, anything else? I guess the surface area is pretty low. My current thinking is that I'll try adding a few more tests, then merge this through to open up to wider feedback.

nblumhardt:
Add: _fix whatever is breaking the Travis build_ to that list :-)

src/Serilog.Sinks.File/Sinks/File/IOErrors.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright 2013-2016 Serilog Contributors
```
tsimbalar: nitpicking : Year is now 2017 :)

src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
```diff
@@ -41,14 +126,20 @@ public static class FileLoggerConfigurationExtensions
...
         /// <param name="fileSizeLimitBytes">The approximate maximum size, in bytes, to which a log file will be allowed to grow.
         /// For unrestricted growth, pass null. The default is 1 GB. To avoid writing partial events, the last event within the limit
         /// will be written in full even if it exceeds the limit.</param>
         /// <param name="buffered">Indicates if flushing to the output file can be buffered or not. The default
         /// is false.</param>
         /// <param name="shared">Allow the log file to be shared by multiple processes. The default is false.</param>
         /// <param name="flushToDiskInterval">If provided, a full disk flush will be performed periodically at the specified interval.</param>
+        /// <param name="rollingInterval">The interval at which logging will roll over to a new file.</param>
+        /// <param name="rollOnFileSizeLimit">If <code>true</code>, a new file will be created when the file size limit is reached. Filenames 
+        /// will have a number appended in the format <code>_NNNNN</code>, with the first filename given no number.</param>
```
tsimbalar: If the file is formatted with int.ToString("000"), I guess it should say "_NNN" here

src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
```diff
@@ -92,6 +189,12 @@ public static class FileLoggerConfigurationExtensions
...
```
tsimbalar: same here, shouldn't it be `_NNN` ?

test/Serilog.Sinks.File.Tests/RollingFileSinkTests.cs
```diff
@@ -0,0 +1,150 @@
...
+            finally
+            {
+                var disposable = (IDisposable)log;
+                if (disposable != null) disposable.Dispose();
+                Directory.Delete(temp, true);
+            }
+        }
+
+        [Fact]
+        public void AssemblyVersionIsFixedAt200()
```
tsimbalar: Does that test pass, considering current File Sink (nuget package) version is at least v3.2.0+ ?

nblumhardt: Thanks for all the feedback!

This one does pass - both sinks have had a frozen assembly version since 2.0.0, as a mitigation against the endless stream of binding redirect issues people hit (and report) otherwise :-)

test/Serilog.Sinks.File.Tests/TemplatedPathRollerTests.cs
```diff
@@ -0,0 +1,104 @@
...
+using Xunit;
+
+namespace Serilog.Sinks.File.Tests
+{
+    public class PathRollerTests
+    {
+        [Fact]
+        public void TheLogFileIncludesDateToken()
+        {
+            var roller = new PathRoller("Logs\\log-.txt", RollingInterval.Day);
```
tsimbalar: Aren't backslashes problematic on Linux ? You should probably use [`Path.Combine`](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.combine?view=netcore-2.0#System_IO_Path_Combine_System_String___) , or maybe [`Path.DirectorySeparatorChar`](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.directoryseparatorchar?view=netcore-2.0) for the tests to work cross-platform

tsimbalar: Also maybe [`Path.GetInvalidFileNameChars`](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getinvalidfilenamechars?view=netcore-2.0) can be useful to avoid generating invalid filenames

tsimbalar: oh well, I see it's covered :) 👍

src/Serilog.Sinks.File/Sinks/File/RollingIntervalExtensions.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            switch (interval)
+            {
+                case RollingInterval.Infinite:
+                    return null;
+                case RollingInterval.Year:
+                    return new DateTime(instant.Year, 0, 0, 0, 0, 0, instant.Kind);
+                case RollingInterval.Month:
+                    return new DateTime(instant.Year, instant.Month, 0, 0, 0, 0, instant.Kind);
+                case RollingInterval.Day:
```
skomis-mm: I see a lot of `ArgumentOutOfRangeException`  coming with the sample project, probably
`new DateTime(instant.Year, 1, 1, 0, 0, 0, instant.Kind);`
`new DateTime(instant.Year, instant.Month, 1, 0, 0, 0, instant.Kind); `
are needed

nblumhardt: Thanks, fixing now 👍

nblumhardt: (Adding some tests....)

test/Serilog.Sinks.File.Tests/TemplatedPathRollerTests.cs
nblumhardt: Thanks for the check; yes, just hacking away at this between other jobs today, missed your comment but have it covered now.

nblumhardt:
Finally made myself get WSL back up and running, instead of working in the dark, think everything should be running smoothly now.

nblumhardt:
README updated: https://github.com/nblumhardt/serilog-sinks-file/blob/af88bceba9ef183e0e9520606541066c2fdf6d89/README.md

Time to push this through to dev? What do you think, folks? :-)

skomis-mm:
Didn't spot any issues more, looks great 👍

nblumhardt:
Thanks @skomis-mm

PR #37, opened 2017-10-27
4.0.0 Release

nblumhardt:
* #35 - roll on year/month/day/hour/minute, file size, or a combination of both
 * Slight modernization of the default output template (only when recompiling against the library; may be a breaking change for some file-based log collectors)


nblumhardt:
image

I think it's unlikely we'll get a lot more feedback on the new features before hitting the button on this one. Seems like a low-risk change for existing (non-rolling) uses of the sink.

nblumhardt:
Folks, any thoughts on this? Will keep the wheels turning and merge later today unless anyone has a review in the pipeline (or needs more time to check it out) :-) Cheers!

PR #65, opened 2018-11-03
add SourceLink support

lrushl:
added support of SourceLink in order to enable better debugging experience. For more info regarding SourceLink please refer to the official repo [here](https://github.com/dotnet/sourcelink)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,7 +6,7 @@ insert_final_newline = true
 indent_style = space
 indent_size = 4
 
-[*.{csproj,json,config,yml}]
+[*.{csproj,json,config,yml,props}]
 indent_size = 2
 
 [*.sh]
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ matrix:
     - os: linux
       dist: trusty
       sudo: required
-      dotnet: 2.0.0
+      dotnet: 2.1.300
       group: edge
 script:
   - ./build.sh
--- a/Build.ps1
+++ b/Build.ps1
@@ -25,9 +25,9 @@ foreach ($src in ls src/*) {
 
     & dotnet build -c Release --version-suffix=$buildSuffix
     if ($suffix) {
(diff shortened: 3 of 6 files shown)


nblumhardt:
Thanks for the PR! Will the beta package reference prevent a stable version of this package being published, or is `PrivateAssets="All"` enough to get this through?

danjagnow:
SourceLink is marked with `<DevelopmentDependency>true</DevelopmentDependency>`, so I believe it won't show up as a transitive dependency for Serilog.

nblumhardt:
Great! Thanks for the replies. Since we'd ideally do this consistently throughout the Serilog org, it seems getting it spot-on here is worthwhile :-)

One last question, in single-published-package repos like the Serilog ones, is there any reason not to include the SourceLink task/references in the CSPROJ files themselves, rather than in a PROPS file? Thanks!

lrushl:
@nblumhardt, you are probably referring to the following [PR](https://github.com/serilog/serilog/pull/1163).

In that PR SourceLink v2 was added. For now, it is being considered as obsolete and SourceLink v3 (which is being added in this PR) would evolve further.  For more details please take a look at [SourceLink and .NET Foundation](https://github.com/ctaggart/SourceLink#net-foundation).

Probably I'd also need to create a PR with this updated version of SourceLink into that repo as well (and other kinds of serilog sinks repos :) heh, a lot of work to do but I hope I'll be able to do it).

Regarding your last question about props vs *PROJ it is a really good question. I'd agree that in single-published-package repos, it probably doesn't give any benefit.
Please let me know if you'd prefer these changes were in PROJ file instead of Directory.Build.props. 

Thank you!

nblumhardt:
Thanks for the reply and details! Yes, I'd forgotten about the original serilog/serilog one; seems like v3 is the way to go, anyway :-)

For consistency, since we don't split e.g. NuGet packaging etc. out into PROPS files, rolling it into the CSPROJ file seems like the way to go 👍

Issue #48, opened 2018-01-30
Support Compression of Rolled Log Files


lajoie:
It would be nice if the File sink supported compressing files as they're rolled out in order to help save on disk space.

nblumhardt:
Thanks for the suggestion! We don't have any plans to attempt this at present - the sink is heavily text-oriented. It would be interesting to see if there are any extension points we could provide that would enable it, but I suspect they would need to be quite intricate.

cocowalla:
If I'm not mistaken, all the actual writing is centralised in `SharedFileSink` - if we put aside rolling for the sake of simplifying this discussion, a compressed output can be achieved by simply wrapping the stream in a `GZipStream` and `BufferedStream`.

A couple of notes:

1. You have no control over when data is flushed to disk - flushing is handled by the GZIP stream
2. In the event of a crash, using compression is still fairly 'safe' - as GZIP is a streaming algorithm, you're not going to lose much data. Recovery is a simple `gunzip < mylog.gz > mylog.recovered.txt`

An alternative is to have some kind of hook for when files are rolled, allowing the application to do something like ZIP the file. But I'm not sure if there is any 'precedent' for hooks in any of the other Serilog sinks? Personally, in any case I'd favour a streaming implementation for its more predictable CPU and disk usage.

cocowalla:
@nblumhardt have you any thoughts on this?

nblumhardt:
Thanks for the nudge! Personally, and this might not be an informed point of view :-), I don't find compressed-file logging appealing. On Windows at least, it's likely easier and more ergonomic just to mark a folder as "compressed" and let the OS/filesystem deal with it, if you're stuck with low-disk-space local file logging.

With disk being at least as cheap to expand as CPU, it seems like adding storage is also probably a better bang-for-the-buck solution.

Adding something like a `Stream`-to-`Stream` extension point is an interesting possibility, though; by plugging in a factory from `Stream` to `Stream`, wrapper streams could be created, and these might be useful for implementing other cases like headers/footers (write header if length == 0 on wrapping, write the footer inside `Dispose()`, etc.), as well as this one.

There are likely quite a few gory details to figure out, though - it'd require some dedicated time and effort. If anyone's interested in enumerating the use cases, how they'd be achieved, and the rough API/internal design, it'd be a worthwhile discussion.

cocowalla:
[...]

PR #80, opened 2019-02-01
`FileLifecycleHooks`

cocowalla:
Enabled wrapping FileSink and RollingFileSink's output stream in another stream, such as a GZipStream. This forms part of a solution to #48


src/Serilog.Sinks.File/StreamWrapper.cs
```diff
@@ -0,0 +1,17 @@
+using System.IO;
+
+namespace Serilog
```
nblumhardt: This would need to be down in `Serilog.Sinks.File` so that we don't pollute the imported namespaces in every file that's `using Serilog;`

src/Serilog.Sinks.File/StreamWrapper.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    /// <summary>
+    /// Wraps the log file's output stream in another stream, such as a GZipStream
+    /// </summary>
+    public abstract class StreamWrapper
```
nblumhardt: Since the parameter lists to methods of are hot real-estate :-) we might avoid some future churn by naming this `FileLifecycleHooks` or something along those lines? It would need some thought put into the naming of `Wrap()`, since the convention would ideally be consistent with future methods we might add, like `OnFileOpened()`, `OnFileClosed()`. (Maybe we could pre-empt some of that? `Wrap()` could _almost_ be the `OnFileOpened()` method, give or take a parameter or two?)

src/Serilog.Sinks.File/StreamWrapper.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        /// <summary>
+        /// Wraps <paramref name="sourceStream"/> in another stream, such as a GZipStream, then returns the wrapped stream
+        /// </summary>
+        /// <param name="sourceStream">The source log file stream</param>
+        /// <returns>The wrapped stream</returns>
+        public abstract Stream Wrap(Stream sourceStream);
```
nblumhardt: Perhaps `sourceStream` here  could be `underlyingStream`? The `<comments>` or `<remarks>` here should probably document the contract for disposing the underlying stream, i.e. if the wrapper applies wrapping, then it's taking on responsibility for disposing the underlying stream when the wrapper is disposed.

src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
```diff
@@ -156,7 +157,8 @@ public static class FileLoggerConfigurationExtensions
             RollingInterval rollingInterval = RollingInterval.Infinite,
             bool rollOnFileSizeLimit = false,
             int? retainedFileCountLimit = DefaultRetainedFileCountLimit,
-            Encoding encoding = null)
+            Encoding encoding = null,
+            StreamWrapper wrapper = null)
```
nblumhardt: The added parameter here will be a binary breaking change, even though calling code will recompile successfully.

There are a couple of options - for `FileSink()`, we should probably add a constructor overload to accept the new parameter, and have the old calls redirect to it.

At the level of these extension methods, it's possible to add an overload also, but the old methods have to be carefully un-defaulted and `[Obsolete]`d so that recompilation targets the new method (thereby giving us some future room to remove the old code). There are some examples of this in this file - see the first two methods in this class. Probably the best option for now, though in some future `5.0` we might start cleaning this up.

src/Serilog.Sinks.File/StreamWrapper.cs
cocowalla: I'm with you on renaming `StreamWrapper` to avoid ctor hell as more hooks are added.

At first I wasn't too sure about `OnFileOpened`, since as-is the `Wrap` method is quite different to a 'standard' event handler. But after thinking a bit, you might have a point. Perhaps, as I guess you are alluding to, we could add a `string path` parameter, which would also have the bonus effect of making it more "general purpose". It would become:

`public abstract Stream OnFileOpened(Stream underlyingStream, string path);`

The idea would be that if you're not interested in changing the stream, you just return `underlyingStream` (this would be in the xml docs)

src/Serilog.Sinks.File/StreamWrapper.cs
cocowalla: Agreed on `underlyingStream`.

Also agreed on the docs, I should have made that clear. Serilog further wraps the wrapped stream with a `StreamWriter`, effectively taking ownership of the wrapped stream.

```
/// <remarks>
/// Serilog is responsible for disposing of the wrapped stream
/// </remarks>
```

src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
cocowalla: Regarding `FileSink`, I just wanted to be clear that this is what you mean:

```
// Original ctor redirects to new one
public FileSink(string path, ITextFormatter textFormatter, long? fileSizeLimitBytes, Encoding encoding = null, bool buffered = false)
	: this(path, textFormatter, fileSizeLimitBytes, encoding, buffered, null)
{
  // Intentionally emtpy
}

// New ctor
public FileSink(string path, ITextFormatter textFormatter, long? fileSizeLimitBytes, Encoding encoding = null, bool buffered = false,
	FileLifecycleHooks hooks = null)
{
  // Code here
}
```
        
After doing that, existing calls such as `new FileSink(path, new JsonFormatter(), null)` result in a compiler error due to `The call is ambiguous`. I can of course change all the calls in the library so they use the new ctor, but I guess this ctor ambiguity will break things for users?

The extension methods are rather fiddly to get right :nerd: but I'll see what I can do!

nblumhardt: Usually in the past we've resolved the ambiguity by removing the `= x` default parameter values from the older version of the constructor. Alternatively, the new parameter can be positioned before the first defaulted one and not given a default value.

cocowalla: If adding a new defaulted parameter breaks binary compatibility, I would have though that removing defaults would have too?

nblumhardt: From the loader's perspective, only the number/kind/order of parameters is important, it doesn't see the default values at all. So adding a parameter, even if it has a default, changes the signature. Adding/removing/modifying default values doesn't break binary compatibility because they're not considered part of the signature. Fun and games :)

cocowalla: Hmm, OK, interesting :smile:

So let's say that a user calls this today:

`new FileSink(path, textFormatter, 12345);`

And then we remove the defaults from the remaining `encoding` and `buffered` parameters, and do this:

```
// Original ctor - defaults removed, and now redirects to new ctor
public FileSink(string path, ITextFormatter textFormatter, long? fileSizeLimitBytes, Encoding encoding, bool buffered)
	: this(path, textFormatter, fileSizeLimitBytes, encoding, buffered, null)
{
  // Intentionally emtpy
}

// New ctor
public FileSink(string path, ITextFormatter textFormatter, long? fileSizeLimitBytes, Encoding encoding = null, bool buffered = false,
	FileLifecycleHooks hooks = null)
{
  // Code here
}
```

In this scenario, will the old code continue to work?

I feel there must be some tooling that can check if binary compatibility is maintained - do you use anything like that?

nblumhardt: Yes, the old code will be binary compatible here, because the compiler inserts the default values directly into the call site.

So far I haven't found/used any recent tooling for this, but it could be out there. Mostly just painstaking work :-)

tsimbalar: https://www.fuget.org/ can be used to compare the API changes between 2 published versions of the same Nuget package. Quite helpful before remove the "pre-release" suffix :)

nblumhardt:
Hi @cocowalla - hope all's going well! I don't think we've synced up since your last round of changes was pushed - is this ready for another look? Cheers!

cocowalla:
@nblumhardt it's a bit of a chore going through the overloads, so I confess I've been putting it off :smile: 

I'd still really like this in Serilog tho, and I believe it should indeed enable #95 too - I'll try to look at it this week!

nblumhardt:
If you'd like, I can try my hand at sending the last few tweaks through to your branch? I should have some time in the next few days :-D

cocowalla:
I didn't get time this week, so that would be great :)

If you don't get the time though, I should have some later next week (hopefully!)

src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
```diff
@@ -309,11 +315,16 @@ public static class FileLoggerConfigurationExtensions
 #pragma warning disable 618
                     if (shared)
                     {
+                        if (hooks != null)
+                        {
+                            SelfLog.WriteLine("Unable to use output stream wrapper - these are not supported for shared log files");
```
tsimbalar: maybe having a message being a bit more detailed could help track down who wrote this `SelfLog` message when investigating issues ? 

I think it would be good if the message at least mentioned `FileLifecycleHooks` to get an idea of what the user is doing wrong.

tsimbalar: Also, if I remember the guidelines of `Serilog` properly, we consider that it's *OK* to throw / crash during configuration of the pipeline when it is a "developer error" (but should not throw later on).

see https://github.com/serilog/serilog/wiki/Reliability#configuration for more.
Probably @nblumhardt has a clearer/more opinionated view on the philosophy :)

tsimbalar:
(Oops, accidentally pressed the `Approve` button .... not sure how I can undo this)

src/Serilog.Sinks.File/Sinks/File/FileSink.cs
```diff
@@ -68,6 +70,11 @@ public FileSink(string path, ITextFormatter textFormatter, long? fileSizeLimitBy
                 outputStream = _countingStreamWrapper = new WriteCountingStream(_underlyingStream);
             }
 
+            if (hooks != null)
+            {
+                outputStream = hooks.Wrap(outputStream);
```
tsimbalar: not sure how paranoid we need to be here ...

Should we be specially careful about cases where : 
1. `.Wrap()` returns null ? 
2. `.Wrap()` throws an exception ?
4. `.Wrap()` returns a `Stream` that is not *writable* ?

In case 1, I guess `new StreamWriter(outputStream)` would end up throwing `ArgumentNullException`. The error message may not be super clear in that case ? 

In case 2, I guess the exception would just bubble up, that's probably good enough

In case 3, `new StreamWriter(outputStream)` would end up throwing an `ArgumentException` which hopefully would be clear enough

src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
cocowalla: > Also, if I remember the guidelines of `Serilog` properly, we consider that it's _OK_ to throw / crash during configuration of the pipeline when it is a "developer error" (but should not throw later on).
> 
> see https://github.com/serilog/serilog/wiki/Reliability#configuration for more.
> Probably @nblumhardt has a clearer/more opinionated view on the philosophy :)

I purposely didn't throw because I remembered Serilog's "no throw" philosophy :)   Easily changed if throwing is allowed at startup tho

cocowalla: > 
> 
> maybe having a message being a bit more detailed could help track down who wrote this `SelfLog` message when investigating issues ?
> 
> I think it would be good if the message at least mentioned `FileLifecycleHooks` to get an idea of what the user is doing wrong.

Very sensible - I just pushed e604418ccaad879556a01b33aa6d34b6dac1c759 for this

src/Serilog.Sinks.File/Sinks/File/FileSink.cs
cocowalla: 1. Agreed, it's going to result in an `ArgumentNullException` later on, so should we just throw our own, clearer exception here? Something like:

```
outputStream = hooks.Wrap(outputStream);

if (outputStream == null)
{
    throw new ArgumentNullException($"{hooks.GetType}.Wrap returned null when wrapping the output stream");
}
```

2. Agreed, I doubt we could add anything useful beyond the originator's exception

3. Agreed, the message would be clear here

tsimbalar: > 1. Agreed, it's going to result in an ArgumentNullException later on, so should we just throw our own, clearer exception here?

sounds good, but I would probably avoid `ArgumentNullException` (which is typically used to validate incoming arguments, for *pre-condition* sort of checks) but more something generic like `InvalidOperationException`.

I think your Exception message is good 👍

cocowalla: Good point - just pushed 873f6d453950e33daf58a6dd38b2a0a60bfb39b6 for this

src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
nblumhardt: Great point @tsimbalar - this is one of the cases (incorrect arguments) where I think we usually throw. (E.g. this could be an `ArgumentException` like the ones above in https://github.com/serilog/serilog-sinks-file/blob/dev/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs#L292 ).

cocowalla: Just pushed e310ab9bd2ab7648d4629b994714240be53a24d6 to throw in this case, instead of just logging. Also inlcluded a test for good measure.

nblumhardt: Awesome 👍 - I will check out the overload situation later today, and with some luck, push this all through to dev! :-)

nblumhardt:
Jumping into this now! :-)

src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
```diff
@@ -287,35 +292,37 @@ public static class FileLoggerConfigurationExtensions
...
                     {
-                        sink = new SharedFileSink(path, formatter, fileSizeLimitBytes);
+#pragma warning disable 618
+                        sink = new SharedFileSink(path, formatter, fileSizeLimitBytes, encoding);
+#pragma warning restore 618
                     }
                     else
                     {
-                        sink = new FileSink(path, formatter, fileSizeLimitBytes, buffered: buffered);
+                        sink = new FileSink(path, formatter, fileSizeLimitBytes, encoding, buffered, hooks);
```
nblumhardt: I am not sure why `encoding` was not propagated, here and a few lines above - I think it's a bug (though it's hard to argue with an enforced UTF-8 no-BOM default :-) )

src/Serilog.Sinks.File/FileLifecycleHooks.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// file headers, or wrap the stream in another that adds buffering, compression, encryption, etc. The underlying
+        /// file may or may not be empty when this method is called.
+        /// </summary>
+        /// <remarks>
+        /// A value must be returned from overrides of this method. Serilog will flush and/or dispose the returned value, but will not
+        /// dispose the stream initially passed in unless it is itself returned.
+        /// </remarks>
+        /// <param name="underlyingStream">The underlying <see cref="Stream"/> opened on the log file.</param>
+        /// <returns>The <see cref="Stream"/> Serilog should use when writing events to the log file.</returns>
+        public virtual Stream OnOpened(Stream underlyingStream) => underlyingStream;
```
nblumhardt: One of the target use cases is writing headers e.g. to CSV/W3C log files, so the name `Wrap()` might be too specific. I've tried `OnOpened()` here, with the possibility of adding `OnOpening(path)` and possibly an `OnClosing(stream)`/`OnClosed(path)` pair later on.

If we make this `virtual` rather than `abstract`, we won't introduce inconsistencies when future hooks are added (since these must be `virtual` to avoid breakage).

All sounds sane?

nblumhardt:
@merbla I'm following your serilog/serilog example to switch from Travis to AppVeyor for Linux builds. Anything special I need to do to get rid of the Travis integration? I removed the webhook but it's back - guessing the Travis project needs to be deleted?

nblumhardt:
Okay, I think this is ready to merge.

There are (as you'd expect) a few tricky design issues.

 * It would be nice to separate wrapping the stream from initializing it with content, i.e. so that a `TextWriter` could be passed to a hook rather than the `Stream` + `Encoding`, but without access to the underlying stream, code that writes headers can't tell whether or not the stream has data yet.
 * Composing hooks is tricky - but this can be punted as the code performing the composition (of say, GZip + Header writer) can itself be a single lifecycle hook that dispatches calls to one or more other hooks.

The AppVeyor build is green; the Travis build is no longer needed.

Any thoughts/comments @cocowalla @tsimbalar , or should we send this through to `dev`? 😄

src/Serilog.Sinks.File/FileLifecycleHooks.cs
cocowalla: Actually, we discussed this a while back in this thread, and it looks like I just forgot to rename this! 

`virtual` reasoning also makes sense.

cocowalla: Back when we started discussing this there was the question of streaming vs 'static' extensibility points and this PR solves the streaming part.

Later bolting on `OnOpening(path)`, `OnClosing(stream)`/`OnClosed(path)` would solve the 'static' part, enabling scenarios such as cryptographically signing files before they are closed, shipping closed files to WORM media, perform (non-streaming) compression and/or encryption etc :+1:

cocowalla:
@nblumhardt looks good to me, thanks!

src/Serilog.Sinks.File/FileLifecycleHooks.cs
nblumhardt: Ha! I actually like the name we were using in the earlier comments a little better, on reflection. I'll rename to `OnFileOpened()` and merge away! :-)

nblumhardt:
Thanks @cocowalla 👍

cocowalla:
@nblumhardt to get things going, I've just published [Serilog.Sinks.File.GZip](https://www.nuget.org/packages/Serilog.Sinks.File.GZip/) as a NuGet package 👍   Looking forward to the release of 4.1.0!

nblumhardt:
@cocowalla that's great! I've put the word out on Twitter, keen to see how it all goes 😊

emiliovmq:
@cocowalla is there any support at Serilog.Sinks.File.Header for DI. Let me explain myself. We have a requirement that dictates that for every log we generate we should add to the beginning of the file the version information of our APIs. As you might expect, I'm reading the version information from the assemblies attributes using reflection (which is a very expensive process). To not incur in this very costly process (for every file event and for my internal requests), I only read the version info once (the first time it is requested), registering it with the DI container and resolve it for every subsequent request or event. l would like to inject the version info into the header writer. Is there any simple workaround for this?

emiliovmq:
I forgot to thank you and @nblumhardt  for this great feature added to Serilog. Thanks

cocowalla:
Just a note to say there is now another FileLifecycleHooks-based plugin available, [serilog-sinks-file-archive](https://github.com/cocowalla/serilog-sinks-file-archive), which works with rolling log files, archiving completed log files before they are deleted by Serilog's retention mechanism.

Issue #64, opened 2018-10-18
Support {Date} for log path


saithis:
We would like to name our log files "appname-{Date}-{ServerId}.txt" ("{ServerId}" is replaced by our code). That way the files could be sorted by date und then by server. But currently we cant decide where the date part should be and it is always appended at the end.

nblumhardt:
Hi! Thanks for the note; this is something supported currently in _Serilog.Sinks.RollingFile_, so at least in the short-term, using that package may be a useful alternative (if you're not reliant on other features not supported, there).

I'd personally like to avoid taking the API down this path, here, because we were never quite able to achieve seamlessness in the way `{Date}` is handled in _RollingFile_ - formatting placeholders don't work as expected, for example, making it a weak analogy with the message/output templates supported elsewhere in Serilog.

One possibility might be to insert a marker character like `*` into the filename, here - e.g.:

```
WriteTo.File("appname-*-foo.txt")
```

which should be effectively non-breaking as `*` would be a very unusual character to see in a log file name, in the wild.

The main things to then determine would be:

 * How would this affect the code that identifies files belonging to the rolling set? (Would prefer that this stays ultra-simple), and,
 * What effect would this have on the numeric file set component - would the above produce `appname-20181019-foo_001.txt` or `appname-20181019_001-foo.txt`, and if so, would this lead to ambiguities when trying to identify files in the rolling set?

These could be non-issues; it'd be great if someone has the time to figure out a solid proposal around this 👍

yhvicey:
What about providing a `Func<string>` as the path generator to `WriteTo.File(...)`? Or some class like `FilePathGenerator` and then we can derive & override it, as an advanced option.

Jehoel:
I've created a PR for this, albiet using a different synax here: https://github.com/serilog/serilog-sinks-file/pull/84

tettsuan:
I have one possible solution. you can use Serilog-Sink-Map to achieve this.
I haven't taken a deep look so the inpact of using that package is not cleared.
```
Log.Logger = new LoggerConfiguration()
    .WriteTo.Map<string>(
         e => DateTime.Now.ToString("yyyyMMdd"),
         (v, wt) => wt.File($"./logs/{v}_log.txt"))
    .CreateLogger();
```

malcolmjerry:
Try this 

 
    private readonly Serilog.ILogger _logger; //= Log.ForContext( "Name", "Weather" );

    public WeatherForecastController() {
[...]

PR #84, opened 2019-02-24
Ability to use DateTime format strings in Rolling File Sinks file names

Jehoel:
This PR adds the ability to use standard Custom .NET DateTime Format Strings when using rolling log files.

This also adds support for using format-strings in subdirectory names, so you can group log files in a directory according to the year, month, or whatever you like.

Also includes tests for the new functionality. All XUnit tests ran in Visual Studio 2017 on Windows 10 pass.

## Errata

There are a few gotchas that probably need addressing before this ships publicly:

- It assumes the code runs in an environment with a case-insensitive file-system (the existing Serilog File Sink code does too). I think it should be aware of case-sensitive file-systems and use different filename comparisons (`StringComparison.Ordinal` vs `StringComparison.OrdinalIgnoreCase` accordingly).
    * However, there is no ready API in .NET Standard, .NET Core or .NET Framework for detecting case-insensitive/case-preserving/case-sensitive filesystems.
- It needs integration tests with the rest of the Serilog system.
- It needs more test-cases for formatted subdirectory names.
- `RollingFileSink::ApplyRetentionPolicy` doesn't delete formatted subdirectories when cleaning up files if the directory becomes empty.
- Some tests fail on Linux (e.g. the current Travis CI results) because the tests use a backslash as a directory separator. What's the best way to fix this so the tests work cross-platform?
    * (.NET Standard _really_ needs a `Path` API so we can stop using `String` for file-names and skirt around these silly cross-platform path issues)
- `PathRoller` probably should be refactored out given the limited value it provides over `IRollingFilePathProvider`. Thoughts?

## Example usage:

The new overloads have two separate string parameters: `logDirectoryPath` and `logFilePathFormat`:

    .WriteTo.File(
        formatter: ...,
        logDirectoryPath: @"C:\Logs\",
        logFilePathFormat: @"yyyy-MM'\Log-'yyyy-MM-dd'.log'",
        rollingInterval: RollingInterval.Day
    )

This will create a log-file named `C:\Logs\2019-02\Log-2019-02-22.log`. If a sequence number is applied then the file-name will be `C:\Logs\2019-02\Log-2019-02-22_001.log`.

The existing overloads that have a single string parameter `path` retain the old behaviour, so this PR does not introduce any breaking changes.

### DIFF
--- a/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
+++ b/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
@@ -172,6 +172,61 @@ namespace Serilog
         /// Write log events to the specified file.
         /// </summary>
         /// <param name="sinkConfiguration">Logger sink configuration.</param>
+        /// <param name="path">Path to the file.</param>
+        /// <param name="pathIsFormatString">If set to <c>true</c>, then <paramref name="path"/> is a DateTime Format String. Use Custom .NET DateTime format specifiers to control how the interval dates are placed in the file-name (including any parent directory names). The sequence number is always inserted before the file-name extension, however.</param>
+        /// <param name="restrictedToMinimumLevel">The minimum level for
+        /// events passed through the sink. Ignored when <paramref name="levelSwitch"/> is specified.</param>
+        /// <param name="levelSwitch">A switch allowing the pass-through minimum level
+        /// to be changed at runtime.</param>
+        /// <param name="formatProvider">Supplies culture-specific formatting information, or null.</param>
+        /// <param name="outputTemplate">A message template describing the format used to write to the sink.
+        /// the default is "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{Level:u3}] {Message:lj}{NewLine}{Exception}".</param>
+        /// <param name="fileSizeLimitBytes">The approximate maximum size, in bytes, to which a log file will be allowed to grow.
+        /// For unrestricted growth, pass null. The default is 1 GB. To avoid writing partial events, the last event within the limit
+        /// will be written in full even if it exceeds the limit.</param>
+        /// <param name="buffered">Indicates if flushing to the output file can be buffered or not. The default
+        /// is false.</param>
+        /// <param name="shared">Allow the log file to be shared by multiple processes. The default is false.</param>
+        /// <param name="flushToDiskInterval">If provided, a full disk flush will be performed periodically at the specified interval.</param>
+        /// <param name="rollingInterval">The interval at which logging will roll over to a new file.</param>
+        /// <param name="rollOnFileSizeLimit">If <code>true</code>, a new file will be created when the file size limit is reached. Filenames 
+        /// will have a number appended in the format <code>_NNN</code>, with the first filename given no number.</param>
+        /// <param name="retainedFileCountLimit">The maximum number of log files that will be retained,
+        /// including the current log file. For unlimited retention, pass null. The default is 31.</param>
+        /// <param name="encoding">Character encoding used to write the text file. The default is UTF-8 without BOM.</param>
+        /// <returns>Configuration object allowing method chaining.</returns>
+        /// <remarks>The file will be written using the UTF-8 character set.</remarks>
+        public static LoggerConfiguration File(
+            this LoggerSinkConfiguration sinkConfiguration,
+            string path,
+            bool pathIsFormatString,
+            LogEventLevel restrictedToMinimumLevel = LevelAlias.Minimum,
+            string outputTemplate = DefaultOutputTemplate,
+            IFormatProvider formatProvider = null,
+            long? fileSizeLimitBytes = DefaultFileSizeLimitBytes,
+            LoggingLevelSwitch levelSwitch = null,
+            bool buffered = false,
+            bool shared = false,
+            TimeSpan? flushToDiskInterval = null,
+            RollingInterval rollingInterval = RollingInterval.Infinite,
+            bool rollOnFileSizeLimit = false,
+            int? retainedFileCountLimit = DefaultRetainedFileCountLimit,
+            Encoding encoding = null)
+        {
+            if (sinkConfiguration == null) throw new ArgumentNullException(nameof(sinkConfiguration));
+            if (path == null) throw new ArgumentNullException(nameof(path));
+            if (outputTemplate == null) throw new ArgumentNullException(nameof(outputTemplate));
+
+            var formatter = new MessageTemplateTextFormatter(outputTemplate, formatProvider);
+            return File(sinkConfiguration, formatter, path, pathIsFormatString, restrictedToMinimumLevel, fileSizeLimitBytes,
+                levelSwitch, buffered, shared, flushToDiskInterval,
+                rollingInterval, rollOnFileSizeLimit, retainedFileCountLimit, encoding);
+        }
+
+        /// <summary>
+        /// Write log events to the specified file.
+        /// </summary>
+        /// <param name="sinkConfiguration">Logger sink configuration.</param>
         /// <param name="formatter">A formatter, such as <see cref="JsonFormatter"/>, to convert the log events into
         /// text for the file. If control of regular text formatting is required, use the other
         /// overload of <see cref="File(LoggerSinkConfiguration, string, LogEventLevel, string, IFormatProvider, long?, LoggingLevelSwitch, bool, bool, TimeSpan?, RollingInterval, bool, int?, Encoding)"/>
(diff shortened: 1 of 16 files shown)


nblumhardt:
Hi! Thanks for the interesting contribution.

This is a rather large PR and has a few subtleties that make it difficult to review as-is without a design document of some kind making clear the capabilities, goals, trade-offs, and edge cases.

For example, does retention delete empty subdirectories? Is there a reason why the directory and path format need to be separate, or could a change in syntax get this back to a single pathTemplate parameter? Do Linux users have to use Windows-style backslashes in paths?

If anyone in @serilog/reviewers-core is able to dig in deeply and bring these things to light, that might help move things forward.

Otherwise I think we'd need to start from the essentials - what are the goals? why approach it this way? does this relate to issues that have been raised and discussed previously? etc., before we'll be able to make progress on the PR.

Jehoel:
In response to @nblumhardt:

> This is a rather large PR and has a few subtleties that make it difficult to review as-is without a design document of some kind making clear the capabilities, goals, trade-offs, and edge cases.

It started off as a quick in-and-out to support `DateTime` format strings in filenames and I polished it over time - so there isn't a design document, but I hope I've provided enough information in this comment reply below to constitute a design document of-sorts:

> does retention delete empty subdirectories?

Presently; No, but _should_ it? (It's an easy change to make)

> Is there a reason why the directory and path format need to be separate,

There are two reasons I went with this design:

1. It means the existing `.WriteTo.File` extension methods can be kept separate from my new extension methods.
2. It means that the `logFilePathFormat` value is (exactly) a `DateTime::ToString( String format )` format string, which has two huge advantages:
    * This means users can use their existing knowledge of Standard and Custom `DateTime` format strings.
    * I didn't use a Composite Format string (e.g. `logFilePathFormat: @"{0:yyyy-MM}\Log-{0:yyyy-MM-dd}.log"`) because Serilog needs to be able to re-parse file-names back to extract their period information and sequence number. _This is not possible with Composite Format strings_, but is possible with `DateTime.ParseExact` by re-using the exact same format string.

> Could a change in syntax get this back to a single `pathTemplate` parameter? 

Possibly, but because it wouldn't be feasible to use a Composite Format string we would have to introduce some custom syntax to denote which part of the string should be extracted and used as the `DateTime` format string.

I did consider this, using a character like `|` to split the path from the formattable part, for example:

    logFilePathTemplate: @"C:\logs\|yyyy-MM'\Log-'yyyy-MM-dd'.log'"

...but Linux allows pretty much every character in a file-name or path, and adding some escape-sequence for `|` (or some other character) would probably be a can of worms with having to handle other edge-cases.

> Do Linux users have to use Windows-style backslashes in paths?

No. Forward-slashes alone work just as fine. I added a test-case for this scenario, which passes, by the way.

> what are the goals?

To allow custom file-name formatting and to allow files to be grouped in subdirectories based on date components that are not part of the rolling interval (e.g. this approach allows directories for each month, even if the rolling interval is daily).

And allowing custom formats in file-names is useful when a human is looking at log filenames - mainly because I want to add some separators between date and time components. Right now my production logfiles look like this:

image

...they'd be easier for me to scan if the names were like `WarningsErrors_2019-03-12.log` instead of `WarningsErrors_20190312.log` - especially if they're rolled on an hourly basis (`2019031211` is not easy to scan).

> why approach it this way? 

As said above, I did consider using Composite Format strings (which would also allow for explicit placement of the sequence number) but Composite Format strings cannot be easily parsed back - so this approach was the simplest to get the desired results.

> does this relate to issues that have been raised and discussed previously? 

Yes. The ability to use custom date formats in log filenames has often been requested, for example:

* https://github.com/serilog/serilog/issues/1123
* https://github.com/serilog/serilog/issues/612
* https://github.com/serilog/serilog-sinks-file/issues/64
* https://stackoverflow.com/questions/44401898/what-file-name-placeholders-are-available-for-serilog
* https://github.com/aarondh/SuperRollingFileAlternate/tree/master/Serilog.Sinks.SuperRollingFileAlternate (this project hasn't been updated in 2 years, but shows that even back then people were looking for custom formats in file-names).

DNF-SaS:
Waiting for this PR to be merged - the **logFilePathFormat** is a great new option.

Jehoel:
For this to be merged:

1. The project owners (@nblumhardt etc) need to agree with this new feature and the current overall design.
2. If they approve, then I'll fix the remaining errata and rebase it on the current `master` branch
3. Then they can merge it

As it's still possible the owners will disagree with this feature, I don't want to spend time fixing the errata unnecessarily.

sylvesp:
Are the new options supported when loading Serilog configuration from AppSettings?

Jehoel:
@sylvesp No. I don't know how to support that but that's definitely something that needs to be added. Is there any documentation on how that can be done?

Here's the checklist of tasks I'll do if this PR's concept is approved by @serilog/reviewers-core :

- [ ] Rebase or remerge with the project
- [ ] Add test-cases for Linux:
    - [ ] Case-sensitive filesystems
    - [ ] Different types of slashes in paths
- [ ] `ApplyRetentionPolicy` should delete empty directories if they were created by the path template
- [ ] Add support for loading configuration from files instead of through the configuration API

sylvesp:
There is a repository for extending AppSettings on windows platform:
https://github.com/serilog/serilog-settings-configuration

nblumhardt:
Hi all! Thanks for everyone's input, here, and the additional info, @Jehoel.

I don't think we want to accept this PR at this time. The overall direction makes a number of implicit compromises in order to use the selected implementation strategy, and I suspect we might do things differently if we looked at this fresh.

For example, reusing .NET DateTime format strings is advantageous in some ways (existing docs, parser, formatter), but it seems to force the separation of the parent directory from the subdirectory format, which wouldn't be necessary if we were to use a template syntax like `pathFormat: "C:\SomeDir\{yyyy}\log-{MM-dd}.txt"`.

 - Our parser for this could extract the un-templated `C:\SomeDir` prefix itself, and
 - Although there isn't a readily-available function to match filenames/extract parameters from them, we could write one.

The .NET DateTime format syntax is also might paint us into a corner. How would the syntax be extended to allow selected placement of numeric filename components like `_001`, for example? Not un-solveable problems, but seems better to consider these kinds of complications from the outset, rather than graft them on later.

Hope this makes sense; if there's a will to get this feature in (which would be great) I think it needs to start from a design proposal and work forward from there, even if that does eventually lead to re-opening this PR.

Since large chunks of `File` are reusable, it might alternatively work to release this as a wrapper sink, e.g. `Serilog.Sinks.TemplatedFile`, in much the same way as `Serilog.Sinks.Map` can be used to insert log event property values into filenames. Taking that route, we'd list it in the Available Sinks wiki page and direct users to it when this feature pops up (we could also link it from the earlier threads).

Thoughts? Apologies in advance if this seems very cautious/conservative - it's not so extreme in other parts of Serilog, but `File` is used _everywhere_, and needs to be essentially unbreakable/highly-backwards-compatible, so the bar for new features is (sometimes a bit painfully) high.

Jehoel:
@nblumhardt I agree with your reasoning entirely.

> Since large chunks of File are reusable, it might alternatively work to release this as a wrapper sink, e.g. Serilog.Sinks.TemplatedFile, in much the same way as Serilog.Sinks.Map can be used to insert log event property values into filenames. Taking that route, we'd list it in the Available Sinks wiki page and direct users to it when this feature pops up (we could also link it from the earlier threads).

I'd be more than happy to go down this route, as it enables users to opt-in to this functionality sooner without committing Serilog to supporting it long-term.

How can we begin?

nblumhardt:
Thanks for the reply! Sorry things are a bit sluggish from here, multitasking as usual :-)

The best option, I think, is to take the fork behind this PR and commit it to a new/fresh repo. After that, and renaming the projects in it, it should be a straightforward process to take a reference to _Serilog.Sinks.File_ and remove all of the duplicated types.

From there, the key's going to be getting it out and tested far and wide - once it's published to NuGet, let us know, and we'll update the docs (and ideally the _Serilog.Sinks.File_ README?) so people can track it down 👍

abrasat:
Will the option to to use DateTime format strings in Rolling File Sinks file names also be available when using rolling files by file size ?
For instance would it be possible to roll max. 30 files of max. 10 MB each, and that the name of each log file contains the time stamp when that file was created?

spiko:
So this will not be merged into serilog.sinks.file? The functionality is available in another package?

Jehoel:
@nblumhardt I'm working on the fork now - I'd like my `Serilog.Sinks.File.TemplatedPath` project to extend `Serilog.Sinks.File` but all of the types are `internal`. I don't want to duplicate all of the code from the `Serilog.Sinks.File` package/assemblies - how should I proceed?

nblumhardt:
Hi @Jehoel - great to hear it's moving forward.

The internals of _File_ have changed quite a bit over time, so some duplication might be preferable to making types public in most cases. Is there any particular type that's a major pain, or a lot of little ones?

nblumhardt:
(Note it's possible to effectively get at the sink constructed by `WriteTo.File()` using the `LoggerSinkConfiguration.Wrap()` method/hack, if the sinks themselves are an issue.)

alrekr:
Hi @Jehoel - any progress?

Plusce:
> Hi @Jehoel - any progress?

Topic bump. It would be great to have this functionality in Serilog.

Jehoel:
@alrekr  @Plusce I haven't put any more time into this since May-June 2019. I don't have anything meaningful to add to this PR yet, unfortunately.

All I can suggest is you fork this repo (`serilog-sinks-file`) and apply this PR locally and see if it even builds...

PR #88, opened 2019-03-12
Check for directory existence before attempting access.

billrob:
Stops exception from being thrown and tripping breakpoints when debugging application code.

Plus good practice to avoid exceptions.


nblumhardt:
Thanks! The check is worthwhile 👍  -- it still needs the `try`/`catch` block in there, though, as the directory could in theory be deleted between the check and the file iteration (unlikely, I know - but there's some pretty defensive code in this sink :-) )

billrob:
I understand and updated.

nblumhardt:
Looks like the Travis build has hung before starting; think this is safe to merge, will keep an eye on later builds from `dev`.

Issue #39, opened 2017-11-09
Date-driven (rather than count-driven) file retention option


stephenpatten:
Nick,

Aside from the fact that I (we) log too much, how do I keep all the log files created per day. It looks like we are only keeping the last 31 files. My thinking would be that with my config that the time period would trump the file size rolling behaviour. i.e. I should have 31 calendar days of logs and each day may have hundreds of files that rolled on size.. Hope that make some sense.  

Below you'll find my logger bootstrap and the packages I'm using.

Thank you,
Stephen 

	var filePath =
		$"{ConfigurationManager.AppSettings["LogFileBaseDir"]}\\ExternalFields_{Environment.MachineName}_.txt";

	const string Workflow_Serilog_V6 = "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz}|{Level:u3}|{CorrelationId}|{MachineName}|{Application}|{ProcessId}|{ThreadId}|{SourceContext:l}|{Message}|{NewLine}{Exception}";

	Log.Logger = new LoggerConfiguration()
		.ReadFrom.AppSettings()
		.Enrich.FromLogContext()
		.Enrich.WithMachineName()
		.Enrich.WithProcessId()
		.Enrich.WithThreadId()
		.Enrich.WithProperty("Application", "ExternalFields")
		/*https://github.com/serilog/serilog-sinks-file*/
		.WriteTo.Async(a => 
				a.File(filePath, 
				outputTemplate: Workflow_Serilog_V6, 
				rollingInterval: RollingInterval.Day, 
				rollOnFileSizeLimit: true, 
				fileSizeLimitBytes: 104857600, 
				buffered: true,
				flushToDiskInterval: TimeSpan.FromMilliseconds(500)))
		.CreateLogger();

<packages>
  <package id="Serilog" version="2.5.0" targetFramework="net452" />
  <package id="Serilog.Enrichers.Environment" version="2.1.2" targetFramework="net452" />
  <package id="Serilog.Enrichers.Process" version="2.0.1" targetFramework="net452" />
  <package id="Serilog.Enrichers.Thread" version="3.0.0" targetFramework="net452" />
  <package id="Serilog.Formatting.Compact" version="1.0.0" targetFramework="net452" />
  <package id="Serilog.Settings.AppSettings" version="2.1.2" targetFramework="net452" />
  <package id="Serilog.Sinks.Async" version="1.1.0" targetFramework="net452" />
[...]

PR #90, opened 2019-03-14
File retention policy by date/time

thiagosgarcia:
Add configuration to discard files older than X time defined by a TimeSpan variable.

As discussed in issue #39, I think (and it did work for me) that would be nice to have a dateTime-driven file retention policy. 
There are some security policies that, for example, demands that all log files should be wiped after a few days or weeks. At the same time, you can have as much as possible data in logs, so, for some cases, the strongest rule for deleting log files would be date, not file count or size.

In the example below, files older than 7 days are deleted by `RollingFileSink.ApplyRetentionPolicy(_)` at a maximum count of 100 files, rolling by day or at a limit of 50MB per file
```
  "Serilog": {
    "Using": [ "Serilog.Sinks.File" ],
    "MinimumLevel": "Debug",
    "WriteTo": [
      { "Name": "Console" },
      {
        "Name": "File",
        "Args": {
          "path": "App_Data\\Logs\\app_.log",
          "rollingInterval": "Day",
          "fileSizeLimitBytes": "52428800",
          "rollOnFileSizeLimit": "true",
          "retainedFileCountLimit": "100",
          "retainedFileTimeLimit": "7.00:00:00", //Deletes files older than 7 days
          "outputTemplate": "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{Level:u3}] {Message:lj}{NewLine}{Exception}"
        }
      }
    ]
  },
```

### DIFF
--- a/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
+++ b/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
@@ -138,6 +138,7 @@ namespace Serilog
         /// including the current log file. For unlimited retention, pass null. The default is 31.</param>
         /// <param name="encoding">Character encoding used to write the text file. The default is UTF-8 without BOM.</param>
         /// <returns>Configuration object allowing method chaining.</returns>
+        /// <remarks>The file will be written using the UTF-8 character set.</remarks>
         [Obsolete("New code should not be compiled against this obsolete overload"), EditorBrowsable(EditorBrowsableState.Never)]
         public static LoggerConfiguration File(
             this LoggerSinkConfiguration sinkConfiguration,
@@ -165,7 +166,7 @@ namespace Serilog
         /// <param name="sinkConfiguration">Logger sink configuration.</param>
         /// <param name="formatter">A formatter, such as <see cref="JsonFormatter"/>, to convert the log events into
         /// text for the file. If control of regular text formatting is required, use the other
-        /// overload of <see cref="File(LoggerSinkConfiguration, string, LogEventLevel, string, IFormatProvider, long?, LoggingLevelSwitch, bool, bool, TimeSpan?, RollingInterval, bool, int?, Encoding, FileLifecycleHooks)"/>
+        /// overload of <see cref="File(LoggerSinkConfiguration, string, LogEventLevel, string, IFormatProvider, long?, LoggingLevelSwitch, bool, bool, TimeSpan?, RollingInterval, bool, int?, Encoding, FileLifecycleHooks, TimeSpan?)"/>
         /// and specify the outputTemplate parameter instead.
         /// </param>
         /// <param name="path">Path to the file.</param>
@@ -181,7 +182,7 @@ namespace Serilog
         /// <param name="shared">Allow the log file to be shared by multiple processes. The default is false.</param>
         /// <param name="flushToDiskInterval">If provided, a full disk flush will be performed periodically at the specified interval.</param>
         /// <param name="rollingInterval">The interval at which logging will roll over to a new file.</param>
-        /// <param name="rollOnFileSizeLimit">If <code>true</code>, a new file will be created when the file size limit is reached. Filenames
+        /// <param name="rollOnFileSizeLimit">If <code>true</code>, a new file will be created when the file size limit is reached. Filenames 
         /// will have a number appended in the format <code>_NNN</code>, with the first filename given no number.</param>
         /// <param name="retainedFileCountLimit">The maximum number of log files that will be retained,
         /// including the current log file. For unlimited retention, pass null. The default is 31.</param>
@@ -227,13 +228,18 @@ namespace Serilog
         /// <param name="shared">Allow the log file to be shared by multiple processes. The default is false.</param>
         /// <param name="flushToDiskInterval">If provided, a full disk flush will be performed periodically at the specified interval.</param>
         /// <param name="rollingInterval">The interval at which logging will roll over to a new file.</param>
-        /// <param name="rollOnFileSizeLimit">If <code>true</code>, a new file will be created when the file size limit is reached. Filenames
+        /// <param name="rollOnFileSizeLimit">If <code>true</code>, a new file will be created when the file size limit is reached. Filenames 
         /// will have a number appended in the format <code>_NNN</code>, with the first filename given no number.</param>
         /// <param name="retainedFileCountLimit">The maximum number of log files that will be retained,
         /// including the current log file. For unlimited retention, pass null. The default is 31.</param>
         /// <param name="encoding">Character encoding used to write the text file. The default is UTF-8 without BOM.</param>
         /// <param name="hooks">Optionally enables hooking into log file lifecycle events.</param>
+        /// <param name="retainedFileTimeLimit">The maximum time after the end of an interval that a rolling log file will be retained.
+        /// Must be greater than or equal to <see cref="TimeSpan.Zero"/>.
+        /// Ignored if <paramref see="rollingInterval"/> is <see cref="RollingInterval.Infinite"/>.
+        /// The default is to retain files indefinitely.</param>
         /// <returns>Configuration object allowing method chaining.</returns>
+        /// <remarks>The file will be written using the UTF-8 character set.</remarks>
         public static LoggerConfiguration File(
             this LoggerSinkConfiguration sinkConfiguration,
             string path,
@@ -249,7 +255,8 @@ namespace Serilog
             bool rollOnFileSizeLimit = false,
             int? retainedFileCountLimit = DefaultRetainedFileCountLimit,
             Encoding encoding = null,
-            FileLifecycleHooks hooks = null)
+            FileLifecycleHooks hooks = null,
+            TimeSpan? retainedFileTimeLimit = null)
         {
             if (sinkConfiguration == null) throw new ArgumentNullException(nameof(sinkConfiguration));
             if (path == null) throw new ArgumentNullException(nameof(path));
@@ -258,7 +265,7 @@ namespace Serilog
             var formatter = new MessageTemplateTextFormatter(outputTemplate, formatProvider);
             return File(sinkConfiguration, formatter, path, restrictedToMinimumLevel, fileSizeLimitBytes,
                 levelSwitch, buffered, shared, flushToDiskInterval,
-                rollingInterval, rollOnFileSizeLimit, retainedFileCountLimit, encoding, hooks);
+                rollingInterval, rollOnFileSizeLimit, retainedFileCountLimit, encoding, hooks, retainedFileTimeLimit);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


thiagosgarcia:
Thanks a lot for your review @nblumhardt 
I'm going to push the suggestions you made as soon as I can.

thiagosgarcia:
@nblumhardt, thanks again for the review and suggestions. I pushed the new code including you suggestions, but I just noticed that I didn't updated my forked code with the new `FileLifecycleHooks` parameter on the constructors. I'm going to fix it later today, then I push again.

src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
```diff
@@ -165,7 +167,7 @@ public static class FileLoggerConfigurationExtensions
             var formatter = new MessageTemplateTextFormatter(outputTemplate, formatProvider);
             return File(sinkConfiguration, formatter, path, restrictedToMinimumLevel, fileSizeLimitBytes,
                 levelSwitch, buffered, shared, flushToDiskInterval,
-                rollingInterval, rollOnFileSizeLimit, retainedFileCountLimit, encoding);
+                rollingInterval, rollOnFileSizeLimit, retainedFileCountLimit, encoding, retainedFileDateLimit);
```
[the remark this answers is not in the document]
thiagosgarcia: Just... beautiful! =)
Thanks!

src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
```diff
@@ -287,20 +291,22 @@ public static class FileLoggerConfigurationExtensions
...
-            int? retainedFileCountLimit)
+            int? retainedFileCountLimit,
+            TimeSpan? retainedFileDateLimit)
         {
             if (addSink == null) throw new ArgumentNullException(nameof(addSink));
             if (formatter == null) throw new ArgumentNullException(nameof(formatter));
             if (path == null) throw new ArgumentNullException(nameof(path));
             if (fileSizeLimitBytes.HasValue && fileSizeLimitBytes < 0) throw new ArgumentException("Negative value provided; file size limit must be non-negative.", nameof(fileSizeLimitBytes));
             if (retainedFileCountLimit.HasValue && retainedFileCountLimit < 1) throw new ArgumentException("At least one file must be retained.", nameof(retainedFileCountLimit));
+            if (retainedFileDateLimit.HasValue && retainedFileDateLimit < TimeSpan.FromSeconds(1)) throw new ArgumentException("TimeSpan-based retention policy must retain files for at least 1 second.", nameof(retainedFileDateLimit));
```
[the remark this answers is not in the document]
thiagosgarcia: Yes, makes much more sense. Did that too.

src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs
```diff
@@ -169,28 +173,40 @@ void OpenFile(DateTime now, int? minSequence = null)
...
+                    .ToList();
+
+            if (_retainedFileDateLimit.HasValue)
+                toRemoveByDate = newestFirst
+                    .Where(n => StringComparer.OrdinalIgnoreCase.Compare(currentFileName, n.Filename) != 0)
+                    .Where(x => x.DateTime.HasValue && DateTime.Now.Subtract(_retainedFileDateLimit.Value).CompareTo(x.DateTime.Value) == 1)
+                    .Select(x => x.Filename)
+                    .ToList();
+
+            foreach (var obsolete in toRemoveByCount.Concat(toRemoveByDate.Except(toRemoveByCount)))
```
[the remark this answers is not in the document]
thiagosgarcia: This was a little tricky. 
I needed to move the second Where() logic to a `SkipWhile()` in order to join both conditions.
The result is now the same as the two previous `Where()`, but in a single enumeration =)

Also added a few test cases covering new possible retaining scenarios (Time / Count).

thiagosgarcia:
Just rebased my changes onto the upstream, fixed conflicts and ran unit and manual tests here.
If there's any other thoughts, I'll be glad to hear.
Cheers!

src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
```diff
@@ -138,6 +138,7 @@ public static class FileLoggerConfigurationExtensions
         /// including the current log file. For unlimited retention, pass null. The default is 31.</param>
         /// <param name="encoding">Character encoding used to write the text file. The default is UTF-8 without BOM.</param>
         /// <returns>Configuration object allowing method chaining.</returns>
+        /// <remarks>The file will be written using the UTF-8 character set.</remarks>
```
nblumhardt: This is not the case for the configuration methods that accept `encoding` (default UTF-8 is documented under that parameter).

src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs
```diff
@@ -173,25 +178,29 @@ void OpenFile(DateTime now, int? minSequence = null)
...
                 .Select(Path.GetFileName)
-                .Union(new [] { currentFileName });
+                .Union(new[] { currentFileName });
 
             var newestFirst = _roller
                 .SelectMatches(potentialMatches)
                 .OrderByDescending(m => m.DateTime)
                 .ThenByDescending(m => m.SequenceNumber)
-                .Select(m => m.Filename);
+                .Select(m => new { m.Filename, m.DateTime });
```
nblumhardt: Is it worth selecting a `FileName` and `DateTime` out, here? Seems like we could just drop this line.

src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs
```diff
@@ -173,25 +178,29 @@ void OpenFile(DateTime now, int? minSequence = null)
...
 
             var toRemove = newestFirst
-                .Where(n => StringComparer.OrdinalIgnoreCase.Compare(currentFileName, n) != 0)
-                .Skip(_retainedFileCountLimit.Value - 1)
+                .Where(n => StringComparer.OrdinalIgnoreCase.Compare(currentFileName, n.Filename) != 0)
+                .SkipWhile((x, i) => (i < (_retainedFileCountLimit - 1 ?? 0)) &&
```
nblumhardt: The default of `0` for a null `_retainedFileCountLimit` seems wrong here, since `i` will never be less than it (`null` is intended to indicate "no limit").

Could pulling this expression out into a local function (with some locals, or structured `return (true|false)` make the logic a bit clearer?

src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs
thiagosgarcia: Actually, DateTime is used down there (line 202) to define wich files will be deleted or not.

src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs
thiagosgarcia: I put `0` there, thinking that if `retainedFileCountLimit` got there as null, then I'll ignore this condition. Do you think maybe `31` (the default value) would be better instead (in case of `null`)?

I'll extract a function to get this piece of code clearer. I also didn't actually like it the way it is.

src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
thiagosgarcia: Ok, I'll remove then. This appeared after I rebased my branch and I ended up ignoring it.

thiagosgarcia:
No worries =)
Gonna push this changes as soon as I can.

src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs
nblumhardt: 👍 yes, but it's already a property on the object being selected from; without this line, the enumerator will still have access to both of these members.

src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
nblumhardt: I think this was removed recently from dev 👍

src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs
nblumhardt: I think it'll be hard to get the logic right here without checking `retainedFileCountLimit` against null, but guessing this will all come out more clearly with some reorganisation :-)

src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs
thiagosgarcia: Uh, you're right man! :sweat_smile: It was a list of string but it isn't anymore. Thanks!

thiagosgarcia:
@nblumhardt I was able to push the requested changes. If there's anything else, please, let me know and I'll correct right away.

I did give some thoughts on that `SkipWhile` filter, and tried to create a more self-explanatory function, hope it's good to ship now. =)

iamandymcinnes:
Just wondered if there is any update on getting this PR merged?

nblumhardt:
Thanks for the nudge, @iamandymcinnes - I'm keen to merge this PR, I need to sanity check myself before pushing the button (there's a binary breaking change which might necessitate downstream changes, too), I'll try to loop back around ASAP, if no one else beats me to it :-)

ericvb:
Same question as @iamandymcinnes, what is the status of this PR?

pnstickne:
Bump. The file-only stock retention policy for files is very limiting. Services that restart many times in a short period of time (and create a new file each time) run the risk of losing logs.

nblumhardt:
@pnstickne @ericvb thanks for the bump; sorry again about the delay, @thiagosgarcia - I'll get there! 😅 

Paul, check out https://github.com/serilog/serilog-sinks-map as one way to address the file-set-per-PID type scenarios (assuming you've tried and ruled out passing `shared: true` to your file sink).

asalhani:
@thiagosgarcia Thank you for your effort to include this as a feature.

@nblumhardt Any update on this PR? Any expected time when it will be closed?

nblumhardt:
Fixed my test breakage locally on the way through the works :-)

nblumhardt:
Out in the 5.0-dev build 🎉  - thanks @thiagosgarcia :-)

thiagosgarcia:
Thank you @nblumhardt! If there's anything else I can do, please let me know =)

ThilinaJ:
Hello, 
With what release will this feature be released on NuGet?

Youssef-Ehab77:
Hello, 
I used retainedFileTimeLimit: TimeSpan.FromMinutes(10), but I'm facing weird behavior, it deletes the old file once it reaches its  fileSizeLimitBytes: 10000 and creates a new file even thought I'm still inside the 10 minutes time frame and my retainedFileCountLimit: 50 and  rollingInterval: RollingInterval.Day. But when I make retainedFileTimeLimit: TimeSpan.FromDays(10) it creates a new file once the file size limit exceeds and it doesn't delete the old file.

PR #95, opened 2019-04-12
New logFileHeaders parameter added to configuration method

kamilozturk:
We want to create a w3c formed special log files. It requires a special #Fields directive before log lines.
If we can do this, we can review our log files via Log Parser Studio tool. So we have created a new logFileHeaders parameter and we have inserted to configuration method to able append header text to the log file after a new log file generated.

### DIFF
--- a/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
+++ b/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
@@ -12,9 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using System;
-using System.ComponentModel;
-using System.Text;
 using Serilog.Configuration;
 using Serilog.Core;
 using Serilog.Debugging;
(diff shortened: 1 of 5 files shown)


nblumhardt:
Hi! I think @cocowalla's #80 will cover this case (write a hook that emits the headers if the new file stream being wrapped is zero-length). It would be great to have your thoughts on this approach. Cheers!

nblumhardt:
Closing as I think we'll initially cover this via #80; thanks again for sending this, hopefully this use case will be covered soon :+1:

nblumhardt:
Quick note, the latest -dev package on NuGet now supports this, though you need to write the hook yourself; there's an example in:

https://github.com/serilog/serilog-sinks-file/blob/dev/test/Serilog.Sinks.File.Tests/FileSinkTests.cs#L187

cocowalla:
@kamilozturk @nblumhardt I've just published a NuGet package, [Serilog.Sinks.File.Header](https://www.nuget.org/packages/Serilog.Sinks.File.Header/), which contains a header-writing implementation - just let me know if you encounter any issues

Issue #106, opened 2019-07-05
Allow user to execute custom action on deprecated files (ex. move or send to different server)


nogard111:
Can we allow different action on obsolete files ?
I think it would be beneficial in situation when old logs should be archived by moving to different folder or different server (dedicated only for logs)

 `     /// Action that should be preformed on obsolete(old) log files` 
`         public static Action<string> OnObsoleteFile = null;`

Inside remove loop:

```
          foreach (var obsolete in toRemove)
            {
                var fullPath = Path.Combine(_roller.LogFileDirectory, obsolete);
                try
                {
                    if(OnObsoleteFile!= null)
                        OnObsoleteFile.Invoke(fullPath);
                    else
                    {
                        System.IO.File.Delete(fullPath);
                    }
                }
                catch (Exception ex)
                {
                    SelfLog.WriteLine("Error {0} while processing obsolete file {1}", ex, fullPath);
                }
            }
```

It is more like example, not working code :)

cocowalla:
It's not *exactly* what you describe, but [FileLifecycleHooks](https://github.com/serilog/serilog-sinks-file/blob/36381f6446d79a0c0c8d7eb1a32aec223d35fc1d/src/Serilog.Sinks.File/Sinks/File/FileLifecycleHooks.cs) (available in the preview version) should cover you here.

It allows you to hook in before a new log file is opened, so you can do things like wrap the output stream in a compressing or encrypting stream, add a custom header etc. But at this point you could also scan the log folder for old files that you want to ship, compress, sign,encrypt, delete or whatever.

You can see an example implementation at [/cocowalla/serilog-sinks-file-header](https://github.com/cocowalla/serilog-sinks-file-header).

nblumhardt:
Sounds like a great candidate for an `OnFileRemoving()` hook. Interested in helping to push this through, if you're keen to PR it, @nogard111 ?

cocowalla:
@nblumhardt presumably this would only apply to `RollingFileSink`?

Do we want the presence of an `OnFileRemoving()` hook to mean "the caller will completely handle retention, Serilog does nothing", or "the caller will do something, then Serilog is responsible for deleting the file"?

In terms of API, we could have:
`public abstract void OnFileRemoving(string filename);`

or maybe
`public abstract bool OnFileRemoving(string filename);`

The idea with the latter being that the `bool` signifies whether Serilog should continue to delete the file as normal.

nogard111:
[...]

PR #107, opened 2019-07-08
Add OnFileRemoving life cycle hook #106

nogard111:
Add OnFileRemoving life cycle hook #106

### DIFF
--- a/src/Serilog.Sinks.File/Sinks/File/FileLifecycleHooks.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/FileLifecycleHooks.cs
@@ -37,17 +37,13 @@ namespace Serilog.Sinks.File
         public virtual Stream OnFileOpened(Stream underlyingStream, Encoding encoding) => underlyingStream;
 
         /// <summary>
-        /// Method called on log files that were marked as obsolete (old) and by default would be deleted.
-        /// This can be used to move old logs to archive location or send to backup server
+        /// Method called on log files that were marked as obsolete (old) and would be deleted.
+        /// This can be used to copy old logs to archive location or send to backup server
         /// </summary>
         /// <remarks>
         /// Executing long synchronous operation may affect responsiveness of application
         /// </remarks>
         /// <param name="fullPath">Log file full path</param>
-        /// <returns>
-        /// Return if Serilog should delete file.
-        /// Warning: returning false and keeping file in same place will result in calling this method again in next scan for obsolete files
-        /// </returns>
-        public virtual bool OnFileRemoving(string fullPath) => true;
+        public virtual void OnFileRemoving(string fullPath) {}
     }
 }
--- a/src/Serilog.Sinks.File/Sinks/File/FileLifecycleHooks.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/FileLifecycleHooks.cs
@@ -35,5 +35,19 @@ namespace Serilog.Sinks.File
         /// <param name="encoding">The encoding to use when reading/writing to the stream.</param>
         /// <returns>The <see cref="Stream"/> Serilog should use when writing events to the log file.</returns>
         public virtual Stream OnFileOpened(Stream underlyingStream, Encoding encoding) => underlyingStream;
+
+        /// <summary>
+        /// Method called on log files that were marked as obsolete (old) and by default would be deleted.
+        /// This can be used to move old logs to archive location or send to backup server
+        /// </summary>
+        /// <remarks>
+        /// Executing long synchronous operation may affect responsiveness of application
+        /// </remarks>
+        /// <param name="fullPath">Log file full path</param>
+        /// <returns>
+        /// Return if Serilog should delete file.
+        /// Warning: returning false and keeping file in same place will result in calling this method again in next scan for obsolete files
+        /// </returns>
+        public virtual bool OnFileRemoving(string fullPath) => true;
     }
 }
--- a/src/Serilog.Sinks.File/Sinks/File/FileLifecycleHooks.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/FileLifecycleHooks.cs
@@ -19,6 +19,7 @@ namespace Serilog.Sinks.File
 {
     /// <summary>
     /// Enables hooking into log file lifecycle events.
+    /// Hooks run synchronously and therefore may affect responsiveness of the application if long operations are performed.
     /// </summary>
     public abstract class FileLifecycleHooks
     {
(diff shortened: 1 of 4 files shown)


src/Serilog.Sinks.File/Sinks/File/FileLifecycleHooks.cs
```diff
@@ -35,5 +35,19 @@ public abstract class FileLifecycleHooks
...
+        /// </summary>
+        /// <remarks>
+        /// Executing long synchronous operation may affect responsiveness of application
+        /// </remarks>
+        /// <param name="fullPath">Log file full path</param>
+        /// <returns>
+        /// Return if Serilog should delete file.
+        /// Warning: returning false and keeping file in same place will result in calling this method again in next scan for obsolete files
+        /// </returns>
+        public virtual bool OnFileRemoving(string fullPath) => true;
```
tsimbalar: would returning an `enum` make it clearer what the outcome will be ? (and allow to add some more cases in the future if we need to).

Something like : 

```cs
public enum OnFileRemovingResult
{
    RemoveFile = 0,
    DoNotRemoveFile = 1,
}
```

... or similar, but with better names :P 

_Naming is hard_ ™️

nblumhardt: I was wondering whether it might be preferable not to bring control flow into this; if the hook implementer needed to delete the file, this would still be fine from the sink's point of view, since it's already expected to be resilient in the face of deletion failures.

nblumhardt: (I.e. the hook method would return `void` and the sink would be unchanged other than calling the hook to advise that the file is expected to be deleted.)

cocowalla: So if a hook is specified, it is entirely responsible for handling the file?

nogard111: 1) I think it will be cleaner to return void and replace Serilog action(delete) entirely.

Or 

2) Rename hook to BeforeRemoveFile - and Serilog will always delete after

nblumhardt: > So if a hook is specified, it is entirely responsible for handling the file, and Serilog's default action (delete) won't be performed?

No; I meant the opposite: the hook would be advised if specified, but the actions performed by Serilog would be identical whether there was a hook or not. (I think this is your (2), @nogard111).

In part, this is because it would be strange for more than the retained file count to actually end up being retained, regardless of the presence of hooks. If the configuration suggests 10 files will be retained, I don't think the role of a hook would be to effectively remove this limit as a side-effect of returning `false`.

RE naming, the `*Removing` (before) `*Removed` (after - although we're not providing it) seems okay to me. At least at first sight I don't get the impression that `OnFileRemoving()` is expected to actually remove the file. I could be forgetting other examples that work this way, though!

nogard111: I totally agree that it should not override retainedFileCountLimit settings.

> the hook would be advised if specified, but the actions performed by Serilog would be identical whether there was a hook or not.
Especially looking on second hook(OnFileOpened) following exactly same pattern. I will update PR accordingly

cocowalla: > If the configuration suggests 10 files will be retained, I don't think the role of a hook would be to effectively remove this limit as a side-effect of returning `false`.

Fair enough, I had thought a usage scenario might be implement a more complex retention policy (I don't personally see that as useful though).

Naming `*Removing` makes sense either way though.

nogard111:
If there is no more issues. Please Merge branch :)

src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs
```diff
@@ -199,11 +199,13 @@ void ApplyRetentionPolicy(string currentFilePath)
                 var fullPath = Path.Combine(_roller.LogFileDirectory, obsolete);
                 try
                 {
+                    if (_hooks != null) _hooks.OnFileRemoving(fullPath);
```
nblumhardt:
```csharp
_hooks?.OnFileRemoving(fullPath);
```

src/Serilog.Sinks.File/Sinks/File/FileLifecycleHooks.cs
```diff
@@ -35,5 +35,15 @@ public abstract class FileLifecycleHooks
...
+
+        /// <summary>
+        /// Method called on log files that were marked as obsolete (old) and would be deleted.
+        /// This can be used to copy old logs to archive location or send to backup server
+        /// </summary>
+        /// <remarks>
+        /// Executing long synchronous operation may affect responsiveness of application
+        /// </remarks>
+        /// <param name="fullPath">Log file full path</param>
+        public virtual void OnFileRemoving(string fullPath) {}
```
nblumhardt: I wonder if `OnFileDeleting()` might actually be better? The filesystem APIs use `Delete()` and not `Remove()`, so unless there's a good reason to indicate some difference, the default terminology should do.

Perhaps for the doc we might say:

_Called before an obsolete (rolling) log file is deleted. This can be used to copy old logs to an archive location or send to a backup server, etc._

Note that by convention Serilog's API docs always use full sentences with periods/full stops.

The point about slow operations is good; since this applies to all current and future hooks, perhaps we could add this to the `<summary>` or `<remarks>` for `FileLifecycleHooks` (at the top of the class instead? E.g.:

_Enables hooking into log file lifecycle events. Hooks run synchronously and therefore may affect responsiveness of the application if long operations are performed._

I think we could call the parameter simply `path`, and note in the doc that it's: `The full path to the file being deleted.`.

src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs
```diff
@@ -199,11 +199,13 @@ void ApplyRetentionPolicy(string currentFilePath)
...
+
                     System.IO.File.Delete(fullPath);
                 }
                 catch (Exception ex)
                 {
-                    SelfLog.WriteLine("Error {0} while removing obsolete log file {1}", ex, fullPath);
+                    SelfLog.WriteLine("Error {0} while processing obsolete log file {1}", ex, fullPath);
```
nblumhardt: _Error while **deleting** ..._ ?

nogard111: It could catch also exception from inside of hook (OnFileDeleting) that is why I renamed message to processing.

src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs
nogard111: I am not very fond of shortening everything, and if that is standard in this project then no problem I updated it.

src/Serilog.Sinks.File/Sinks/File/FileLifecycleHooks.cs
nogard111: Good points . Applied

src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs
nblumhardt: 👍

nblumhardt:
Thanks for the updates - this is looking good!

Did you have a chance to investigate how this might be tested? Here's an example of the other existing hook being tested: https://github.com/serilog/serilog-sinks-file/blob/dev/test/Serilog.Sinks.File.Tests/FileSinkTests.cs#L183 - and here's an example of a rolling test that deals with multiple file deletion: https://github.com/serilog/serilog-sinks-file/blob/dev/test/Serilog.Sinks.File.Tests/RollingFileSinkTests.cs#L56

nogard111:
Test added

nblumhardt:
Great! Thanks for the contribution 👍 

This should be out as a `-dev` package soon.

Issue #131, opened 2020-02-09
RollingFileSink _currentFileSequence overflow leads to an infinite loop


trionia:
There is a log file that is going to roll, and its name suffix is Int32.MaxValue, a logger goes into an infinite loop.
When the log file reaches it's size limit, `AlignCurrentFileTo` method gets called with an argument `nextSequence=true`.
[RollingFileSink.cs#L106](https://github.com/serilog/serilog-sinks-file/blob/a015038cdba0a38caa4feb73a78bb85ac0a3464b/src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs#L106)
```cs
                int? minSequence = null;
                if (nextSequence)
                {
                    if (_currentFileSequence == null)
                        minSequence = 1;
                    else
                        minSequence = _currentFileSequence.Value + 1;
                }
```
When `_currentFileSequence.Value` is `Int32.MaxValue`, adding 1 to it results in an overflow, so `minSequence` becomes `Int32.MinValue`.
[RollingFileSink.cs#L142](https://github.com/serilog/serilog-sinks-file/blob/a015038cdba0a38caa4feb73a78bb85ac0a3464b/src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs#L142)
```cs
            var sequence = latestForThisCheckpoint?.SequenceNumber;
            if (minSequence != null)
            {
                if (sequence == null || sequence.Value < minSequence.Value)
                    sequence = minSequence;
            }

```
Later, the second if statement fails on the condition `sequence.Value < minSequence.Value`; `sequence` remains with its old value, file path remains the same. Since the file reached it's size limit `AlignCurrentFileTo` method is getting called again with `nextSequence=true` and the cycle repeats.

Having two billion rolls is rather unlikely to happen, I was just playing around with the library and thought "what if".

cocowalla:
Hah, nice find!

Would be great if you could have a stab at a PR for a failing test for this bug? Something like `FilesShouldContinueRollingOnOverflow`

PR #132, opened 2020-02-11
Added a test for rolling a file when its sequence reached int.MaxValue.

trionia:
A test for an issue https://github.com/serilog/serilog-sinks-file/issues/131.

I specify a file name for a logger to write to, then manually create a file that has the name of the original file with an appended suffix of int.MaxValue. Later, log.Write() will fall into an infinite loop. I didn't figure out how to properly check for it. Also I'm not sure if the rolled file name should be with or without the suffix.

### DIFF
--- a/test/Serilog.Sinks.File.Tests/RollingFileSinkTests.cs
+++ b/test/Serilog.Sinks.File.Tests/RollingFileSinkTests.cs
@@ -128,7 +128,7 @@ namespace Serilog.Sinks.File.Tests
                     Assert.True(System.IO.File.Exists(files[2]));
                 });
         }
-      
+
         [Fact]
         public void WhenRetentionCountAndArchivingHookIsSetOldFilesAreCopiedAndOriginalDeleted()
         {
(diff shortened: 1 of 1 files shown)


trionia:
Well, seems I screwed up. I didn't think about CI running the test. This might be bad. I'm new to contributing, please forgive me :(

trionia:
Should I reopen the PR with a test case that throws an exception?

nblumhardt:
Normally the PR would include the test and fix together - the test alone usually won't be merged unless it passes. HTH, and thanks for contributing!

cocowalla:
@trionia 

> I specify a file name for a logger to write to, then manually create a file that has the name of the original file with an appended suffix of int.MaxValue. Later, log.Write() will fall into an infinite loop

It wasn't immediately obvious to me how to write a test for this, without literally writing `Int32.MaxValue` files to disk - starting with a single file with a high sequence number is a clever approach!

It would be great if you could make the test a bit more CI-friendly (throwing an exception insteadof *actually* letting an infinite loop run away 😉), and then submit a new [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/).

A failing test is just what we need to keep the conversation going on #131, and what we need to test possible solutions.

trionia:
@cocowalla 
I tried to wrap the call that leads to the loop into a Task, but couldn't figure out how to cancel it on a timeout. Even if the test case method itself returns, the task continues running preventing the test from completion. For now the best I have is throwing NotImplementedException [(link)](https://github.com/trionia/serilog-sinks-file/blob/ee3028c38fbbf0fe616916cfa35cdaf598bf9c7c/test/Serilog.Sinks.File.Tests/RollingFileSinkTests.cs#L315). I think its not enough for a draft PR.
I thought about an actual solution of the initial issue. If we could somehow mark existing files with a property like "overflowed by sequence", when the overflow happens, reset the sequence number and create a file from it. Later we could use that property for ordering, so not overflowed files will come before (or after, depending or the orderBy direction) overflowed ones.
But marking files looks too tricky for me. Storing all overflowed file names can be expensive. So maybe the best option is just to rise an exception on overflow and let the end user do handle it.

cocowalla:
@trionia 

There is an overload of `CancellationTokenSource` that takes a `TimeSpan`, such that the token will automatically be cancelled when the time is up.

trionia:
@cocowalla 
The callable code does not support CT. Without modifying the source code of the RollingFileSink it's not possible to stop via cancellation, or at least that's what I've concluded. 
I tried to capture the Thread on which the loop is running and then abort that thread. It worked on .Net 4.52 platform but didn't work on any .Net Core.

cocowalla:
I see what you mean, it's not straightforward to test this. We could run a separate process and kill it, but that really feels like overkill (pun intentional :)

I'll try and take a look at this tomorrow and see if there is any *practical* way to go at this, otherwise we find another way at it.

PR #141, opened 2020-04-17
Add extension method to chain FileLifecycleHooks together

EamonHetherton:
Allows composing of FileLifecycleHooks e.g. Add FileHeaders and GZip the logs, or compress then encrypt.


nblumhardt:
Looks good; since `FileLifecycleHooks` is defined in this assembly, it could be implemented as an instance method rather than an extension?

Since there's a strong concept of ordering between chained hooks, perhaps we could use a name like `Then()` or something along those lines?

```
var hooks = new GZipHooks().Then(new HeaderWriter("File Header"));
```

EamonHetherton:
yeah, agree on both.  Was never happy with "ChainTo" :)

And it is an extension as it started life in my own project, but makes sense as an instance.

will update...

nblumhardt:
LGTM! :+1:

Issue #128, opened 2020-02-04
Recreate current rolling file to log on file deleting


asset-2:
It is not a new topic. It partially interacts with [#36](https://github.com/serilog/serilog-sinks-rollingfile/issues/36), [#87](https://github.com/serilog/serilog-sinks-file/issues/87), [#96](https://github.com/serilog/serilog-sinks-file/issues/96).

Service uses an hour as rolling interval. The file has been deleted in-between. Logger waits till chosen interval ends and then only creates new file to log things. All events and messages are ignored during remained time. Expected behavior for me is to recreate file **immediately** and write to it, as soon as we have something to log.

Do you guys plan to implement this thing? What possible workarounds do you suggest?

cocowalla:
Do you mean you are manually deleting the log file during the period that Serilog would expect to be able to write to it?

asset-2:
Yes, it's possible when you run your application in Docker container for example.

cocowalla:
Hmm, I'm not getting how running in a container deletes log files during writing - would appreciate if you could describe a scenario when this happens?

asset-2:
No, 'running in container' does not delete log files during writing.
Application is running in Docker container. It performs logging into file. It's possible to get access to that log file and delete it manually. I was expecting that file will be recreated immediately, as I wrote before. But no. It creates a new one when rolling interval is over. In this scenario we lose a bunch of messages.

cocowalla:
> It's possible to get access to that log file and delete it manually

OK, but why would you do that? :smile:

I'm trying to figure out if this is a realistic, common thing to happen, or an edge case where you have to actively be trying to cause a problem for it to be an issue. 

Also, as an aside, if you configure the sink using `shared: false`, it will hold an exclusive lock on the log file during writing, preventing you from deleting it.

asset-2:
> Also, as an aside, if you configure the sink using shared: false, it will hold an exclusive lock on the log file during writing, preventing you from deleting it.

I will give a try. But something tells me that this approach does not work.

> OK, but why would you do that? 😄

Someone accidentally removes log file on the production... The data has been lost. At that point we have embarked investigation.
As you see, it is pretty realistic occasion. From my perspective it is pretty evident to logger to proceed without interruption. The client loses its upcomming data for remained rolling period when human/system deletes log file unintentionally or not . Well, I really do not know whether that thing fits well with the strategy that you guys keep.

cocowalla:
[...]

Issue #96, opened 2019-04-17
Allow Log File to be Deleted at Runtime


rustbomber:
I use sinks-file on windows system, when web app runing, log write to file, now I want to delete the log file at runtime, windows report:

File is in use 
The operation could not be completed because the file was opened in dotnet.exe

I use shared parameter, set true, is not working.

my code :
```
const string template = "{Timestamp:yyyy-MM-dd HH:mm:ss.ffffff} {Level:u4} {SourceContext} {Message} {Exception}{NewLine}";
Log.Logger = new LoggerConfiguration()
    .ReadFrom.Configuration(config)
    .Enrich.FromLogContext()
    .MinimumLevel.Is(minimumLevel)
    .WriteTo.Console(outputTemplate: template)
    .WriteTo.Debug(outputTemplate: template)
    .WriteTo.File(path: Path.Combine(AppContext.BaseDirectory, "Logs", "log.txt"), outputTemplate: template,
        fileSizeLimitBytes: 5242880, rollingInterval: RollingInterval.Day, rollOnFileSizeLimit: true, retainedFileCountLimit: 31, shared:true, buffered: false).CreateLogger();
```

nblumhardt:
Hi - `shared: true` sets the file to allow read-write sharing; deleting the file is not supported in this mode.

It seems like there are possibly two separate requirements here:

 1. Allow the file to be deleted
 2. Recreate the file when this happens

I think we could allow (1) without initially worrying about (2) - this might be handy, for instance, in the situation where unbounded log file growth occurs and the file needs to be removed before the process can be (safely?) stopped.

I'm not sure how widely valued this scenario is, though - any thoughts?

lscorcia:
+1 for this issue and the related #128 . I have migrated a large solution from log4net to Serilog only to discover this issue when moving to production. Back to log4net until this gets resolved.

dgxhubbard:
This is definitely an issue I wanted to switch to serilog but found this issue. When the app is allowing the log file to be deleted means there is less to sift through to find a problem. We do this all the time and it saves time finding issues. Why are you locking the log file?

nblumhardt:
Thanks for the nudge!

Nothing's happened here because between my:

> I'm not sure how widely valued this scenario is, though - any thoughts?

and any response was four years, which somewhat slowed thing down :-)

[...]

PR #144, opened 2020-04-28
Ability to delete log files and recreate them immediately

thiagosgarcia:
As discussed in #96 and #128, this gives the ability to delete files when is created by a SharedFileSink and recreates the rolling file when deleted.
Please, note that this is still a initial version for the solution, I'm going to implement more tests and eliminate some duplication. 

I'd like to start this discussion in order gather some insights from the project maintainers and address some questions I have. I'm gonna give some more thoughts on this one as soon as I can.

## Allow Log File to be Deleted at Runtime #96 
I was able to make this work by just adding `FileShare.Delete` at both  `FileStream` instances. This would make all shared files 'deleteable'.

About that: 
- Is ok to make all shared files 'deleteable'?; or
- Should I make another configuration and only enable deletion capability when is configured by the user? (eg.: `canDeleteFiles: true`) 

## Recreate current rolling file immediately to log on file deleting  #128  
This was a little more tricky, but I was able to make this work using `SharedFileSink.OSMutex`

Questions I have:
- How do I reproduce/run `SharedFileSink.AtomicAppend`? I wasn't able to run this class in my tests or by debugging.
- Should we be worried about the performance of `File.Exists()` at `SharedFileSink.OSMutex.cs` line 85? Do you guys have a better idea to make this check? This was actually how I managed to make the file be re-created right after deletion.

### DIFF
--- a/src/Serilog.Sinks.File/Sinks/File/SharedFileSink.AtomicAppend.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/SharedFileSink.AtomicAppend.cs
@@ -72,7 +72,7 @@ namespace Serilog.Sinks.File
                 path,
                 FileMode.Append,
                 FileSystemRights.AppendData,
-                FileShare.ReadWrite,
+                FileShare.ReadWrite | FileShare.Delete,
                 _fileStreamBufferLength,
                 FileOptions.None);
 
@@ -101,7 +101,23 @@ namespace Serilog.Sinks.File
                             _path,
                             FileMode.Append,
                             FileSystemRights.AppendData,
-                            FileShare.ReadWrite,
+                            FileShare.ReadWrite | FileShare.Delete,
+                            length,
+                            FileOptions.None);
+                        _fileStreamBufferLength = length;
+
+                        oldOutput.Dispose();
+                    }
+
+                    if (!System.IO.File.Exists(_path))
+                    {
+                        var oldOutput = _fileOutput;
+
+                        _fileOutput = new FileStream(
+                            _path,
+                            FileMode.Append,
+                            FileSystemRights.AppendData,
+                            FileShare.ReadWrite | FileShare.Delete,
                             length,
                             FileOptions.None);
                         _fileStreamBufferLength = length;
--- a/src/Serilog.Sinks.File/Sinks/File/SharedFileSink.OSMutex.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/SharedFileSink.OSMutex.cs
@@ -30,8 +30,9 @@ namespace Serilog.Sinks.File
     [Obsolete("This type will be removed from the public API in a future version; use `WriteTo.File(shared: true)` instead.")]
     public sealed class SharedFileSink : IFileSink, IDisposable
     {
-        readonly TextWriter _output;
-        readonly FileStream _underlyingStream;
+         TextWriter _output;
+         FileStream _underlyingStream;
+        readonly string _path;
         readonly ITextFormatter _textFormatter;
         readonly long? _fileSizeLimitBytes;
         readonly object _syncRoot = new object();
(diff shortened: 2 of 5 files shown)


src/Serilog.Sinks.File/Sinks/File/SharedFileSink.AtomicAppend.cs
```diff
@@ -72,7 +72,7 @@ public SharedFileSink(string path, ITextFormatter textFormatter, long? fileSizeL
                 path,
                 FileMode.Append,
                 FileSystemRights.AppendData,
-                FileShare.ReadWrite,
+                FileShare.ReadWrite | FileShare.Delete,
```
thiagosgarcia, line 75: File deletion was enabled by adding this, but I don't know if it's the best way to do this.
Should we make another config entry to only enable deletion when required?

src/Serilog.Sinks.File/Sinks/File/SharedFileSink.OSMutex.cs
```diff
@@ -30,8 +30,9 @@ namespace Serilog.Sinks.File
     [Obsolete("This type will be removed from the public API in a future version; use `WriteTo.File(shared: true)` instead.")]
     public sealed class SharedFileSink : IFileSink, IDisposable
     {
-        readonly TextWriter _output;
-        readonly FileStream _underlyingStream;
+         TextWriter _output;
+         FileStream _underlyingStream;
+        readonly string _path;
```
thiagosgarcia, line 35: I needed to make `_output` and `_underlyingStream` non-readonly, so I could dispose them and recreate the file after it was deleted.
Also added `_path` variable to be re-used when recreating the file.

src/Serilog.Sinks.File/Sinks/File/SharedFileSink.OSMutex.cs
```diff
@@ -81,6 +82,13 @@ bool IFileSink.EmitOrOverflow(LogEvent logEvent)
 
                 try
                 {
+                    if (!System.IO.File.Exists(_path))
+                    {
+                        _underlyingStream.Dispose();
+                        _underlyingStream = System.IO.File.Open(_path, FileMode.Append, FileAccess.Write, FileShare.ReadWrite | FileShare.Delete);
+                        _output = new StreamWriter(_underlyingStream,  new UTF8Encoding(encoderShouldEmitUTF8Identifier: false));
+                    }
```
thiagosgarcia, line 90: Although I'm worried about the performance of this `File.Exists()`, this was how I managed to make the file be re-created. If you guys have another idea of how to make this verification, please let me know.
I also noticed that this chunk **must be** before `_underlyingStream.Seek(0, SeekOrigin.End);` in order to create the file correctly after is deleted.

thiagosgarcia:
I noticed that the tests are not passing with VS2019 image. I'm going to check this as soon as I can.

amoerie:
@thiagosgarcia @nblumhardt Any update on this issue? Do you need help?

thiagosgarcia:
Hey! I'm sorry, I've been busy these last months and completely forgot about this.
I recall that I did have some design concerns (written in description), but I'm not quite sure if it still stands. I'm gonna update the code, re run everything and address any question I have in the next few days :)

amoerie:
That's great to hear. I'm willing to help out where I can, but I'm new to the Serilog codebase so I can't promise anything but time and goodwill. 😉

nblumhardt:
Hey all! Looks like this PR has stalled; #186 might offer a partial solution as long as you're not using the `shared` version of the sink?

Doing some housekeeping and clearing out the list of open PRs to keep things manageable; if you do decide to pick it back up, @thiagosgarcia, please just let me know and we can reopen 👍

ShitalPawarCC:
This is a much needed feature. Can someone fix this issue and merge the PR after testing?

thiagosgarcia:
@nblumhardt Completed this implementation in #365 =)
When you have some bandwidth, let me know your thoughts.

PR #164, opened 2020-07-22
Optimization - Use faster null checking

rafaelsc:
**What issue does this PR address?**
None

**Does this PR introduce a breaking change?**
No

**Please check if the PR fulfills these requirements**
- [x] The commit follows our [guidelines](https://github.com/serilog/serilog-sinks-console/blob/dev/CONTRIBUTING.md)
- [ ] Unit Tests for the changes have been added (for bug fixes / features)

**Other Information**:
- Micro-optimization: Replace **null** checking to use `is null` to a 'faster' comparison. _(Possible a very insignificant improvement)_ 
- Code Improvement: Add readonly to the static SyncRoot


cocowalla:
I'm possibly missing something, but wouldn't `is null` only be (potentially) faster if `==` was overloaded?

rafaelsc:
Yes, @cocowalla you are completed right, In the latest Roslyn version is and == will have the only difference is when equals and/or == are overloaded. Like I said: **_(Possible a very insignificant improvement)_.**
This will improve the readability of the code more than the speed.

But testing the same changes in Serilog (core) code base, I see more than **1%** of improvement, and I still don't know why.

cocowalla:
First, let me say that I'm a performance junkie, and spending hours on micro-optimisations is one of my favourite things to do, so thanks for the PR :nerd_face`

From memory, I don't think we overload `==` anywhere, and 1% is well within the typical margin of error for general benchmarks.

From a readability perspective, I don't personally feel strongly either way, and I'm not sure if it's possible to enforce one style over the other using `.editorconfig` (so if we change it, `==` is very likely to creep in again later!).

Since this boils down to code style (personal preference), I'm going to defer to the venerable @nblumhardt on this one :smile:

nblumhardt:
Hi! I'm okay with this change, and would be happy enough to merge, but I'm also unsure whether over time we'll end up with inconsistencies, unless VS/ReSharper/Rider start generating the `is null` form in refactorings etc.

So, +0.5 from me :-)

nblumhardt:
I think for this one we may be better off passing, since we'll otherwise set the expectation for a coding style which may not have any actual benefit. Thanks for the PR though, @rafaelsc 👍

PR #182, opened 2020-10-12
Added Week to the RollingInterval enum.

camerontbelt:
Added a test for Week in RollingIntervalExtenstionsTests class.

### DIFF
--- a/src/Serilog.Sinks.File/RollingInterval.cs
+++ b/src/Serilog.Sinks.File/RollingInterval.cs
@@ -35,6 +35,11 @@ namespace Serilog
         Month,
 
         /// <summary>
+        /// Roll every calendar week. Filenames will have <code>yyyyMMdd</code> appended.
+        /// </summary>
+        Week,
+
+        /// <summary>
         /// Roll every day. Filenames will have <code>yyyyMMdd</code> appended.
         /// </summary>
         Day,
--- a/src/Serilog.Sinks.File/Serilog.Sinks.File.csproj
+++ b/src/Serilog.Sinks.File/Serilog.Sinks.File.csproj
@@ -10,7 +10,7 @@
     <SignAssembly>true</SignAssembly>
     <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
     <PackageTags>serilog;file</PackageTags>
-    <PackageIcon>http://serilog.net/images/serilog-sink-nuget.png</PackageIcon>
+    <IconUrl>http://serilog.net/images/serilog-sink-nuget.png</IconUrl>
     <PackageProjectUrl>http://serilog.net</PackageProjectUrl>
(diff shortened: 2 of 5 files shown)


nblumhardt:
Hi @camerontbelt - thanks for the PR!

I'm not sure this will work - if you shut down the app, and start it up a day later, the new file will be dated only one day later than the previous one, won't it? (The format determines the stride, and not  GetNextCheckpoint(), IIRC?)

I'd also expect that if we're working in "calendar weeks" then we'd pick a week-start day (Sunday, or Monday), but I can't see how that happens so perhaps this is better described as rolling every "seven days"?

camerontbelt:
@nblumhardt Re: every seven days
I definitely think that's pretty much all I was going for. For our needs rolling to the next log once a day is too often but monthly is probably not often enough. I wanted something in between. I know this doesn't take into account calendar weeks, just purely seven days, that's all I was after.

tkrafael:
Guys,
People have all sort of rolling strategies.
Mine is to roll in periods of 5, 10, 15 and 30 minutes.
Do you guys have any iniciative to expose a rolling strategy just like the FileHook class?
I would like to support this change and allow others to configure in more customized ways.

Parsa2820:
> 
> 
> Guys,
> People have all sort of rolling strategies.
> Mine is to roll in periods of 5, 10, 15 and 30 minutes.
> Do you guys have any iniciative to expose a rolling strategy just like the FileHook class?
> I would like to support this change and allow others to configure in more customized ways.

I'm 100 percent agree with that. We need to specify rolling interval more accurate than just `Minute`, `Hour`, `Day`, `Month`, `Year`, and `Infinite`. How about adding a new parameter besides current `RollingInterval`  enum parameter, for example `string preciseRollingInterval` which can get values like `"2h 30m 4s"`?

nblumhardt:
Hi all,

Currently - not planning further work in this area; help is always welcome, but, because of limited reviewer time available, changes to this sink would need to come with a full proposal including analysis of corner cases and other impacts, with time taken to understand the existing codebase, before we could commit to working through a PR.

I'm not suggesting anyone is obliged to do the spec work - but at the moment we can't really move forward on things here.

Parsa2820:
I'm looking forward to work on this feature. Currently, I have a little time to work on it, and it might take a long time. By the way I have created the corresponding issue https://github.com/serilog/serilog-sinks-file/issues/213. If you have any suggestions, I appreciate your comment on the issue.

tkrafael:
I've added a PR that adds support for more granular intervals:
https://github.com/serilog/serilog-sinks-file/pull/205
My scenario is to use Serilog as TDR (transaction detail record) and auditing. Too coarse interval leads to delayed data import, so I need at least 30m intervals.

nblumhardt:
Closing as it seems like this one has stalled on [some limitations](https://github.com/serilog/serilog-sinks-file/pull/182#issuecomment-708050271) - thanks again, @camerontbelt 👍

Issue #186, opened 2020-11-04
Allow FileLifecycleHooks to change the length of the stream


augustoproiete:
Currently files are opened with `FileMode.Append`, which prevents `FileLifecycleHooks` from modifying the size of the stream when implementing an `OnFileOpened` hook.

In addition, when `fileSizeLimitBytes` is set, the stream is wrapped via `WriteCountingStream` which does not implement `SetLength`.

Changing the size of the stream can be useful in a few scenarios, such as:

- [Creating a new log file every time the app runs](https://stackoverflow.com/questions/64657367/is-it-possible-to-configure-serilog-to-truncate-i-e-make-empty-the-log-file-f) (i.e. truncate instead of append)
- Set an initial size to pre-allocate/reserve disk space for the log file (e.g. 1 GB)

---

- Calling `SetLength` with `fileSizeLimitBytes` to `null` throws `System.IO.IOException: Unable to truncate data that previously existed in a file opened in Append mode`

- Calling `SetLength` with `fileSizeLimitBytes` to any number throws `System.NotSupportedException: Specified method is not supported`

PR #187, opened 2020-11-04
Allow FileLifecycleHooks to change the length of the stream

augustoproiete:
- Open files with `FileMode.OpenOrCreate` and immediately position the cursor at the end of the file, effectively doing what the `File` class does behind the scenes when using `FileMode.Append`

- Implement `SetLength` in `WriteCountingStream` adjusting `CountedLength` if necessary (i.e. when the new size of the stream is shorter)

---

Resolves… #186


nblumhardt:
:+1:

PR #189, opened 2020-11-09
Add ability to capture path of log file opened, via FileLifecycleHooks

### LINKED ISSUE #191: Add ability to get the file path of the log file via hooks
Over time I've seen a few questions on StackOverflow asking for a feature similar to what's being asked on https://github.com/serilog/serilog/issues/1489 which is the ability to inspect the properties of the sinks configured, at run-time.

That feature is a bigger piece of work that requires design, of course, however in many of those questions, devs were aiming to use such a feature just to get a hold of the path of the file being written by this sink in particular, for different reasons (\*).

I propose adding the log file path to the existing `OnFileOpened` hook, which would allow devs to plug in a hook that captures the file path of every log file written and keep track of the most recent one.

This would provide a workaround for this particular scenario until https://github.com/serilog/serilog/issues/1489 or something similar becomes available.

Also, for consistency, the `OnFileDeleting` hook already exposes the file path, and I couldn't think of a reason why `OnFileOpened` wouldn't do the same.

---

(\*) I can remember of a couple of examples where devs would like to know the exact file name (considering the name changes with rolling) in order to do run some activity:

- [Get the path to the log file without having to read the app's configuration file, expand environment variables, etc](https://stackoverflow.com/questions/61143684/how-to-get-the-serilog-log-path-from-application-configuration-file)
- [Show a message to the end-user on a desktop app showing where the log file (or folder) is](https://stackoverflow.com/questions/64579477/get-current-serilog-configuration-to-know-full-write-to-file-path)
- [Email the most recent log file to someone before an app closes](https://stackoverflow.com/questions/61317806/email-the-log-file-created-at-the-end-of-program-using-serilog)

augustoproiete:
Adds an overload to `FileLifecycleHooks` that includes the path to the log file that has been opened.

```csharp
public virtual Stream OnFileOpened(string path, Stream underlyingStream, Encoding encoding) { ... }
```

By default, it calls the current `OnFileOpened(Stream, Encoding)` to maintain compatibility with existing hooks.

We can consider marking the current `OnFileOpened(Stream, Encoding)` as `Obsolete` on this or on a future pull-request depending on the next milestone/release.

---

Resolves #191

### DIFF
--- a/test/Serilog.Sinks.File.Tests/Support/CaptureFilePathHook.cs
+++ b/test/Serilog.Sinks.File.Tests/Support/CaptureFilePathHook.cs
@@ -1,0 +1,20 @@
+﻿using System.IO;
+using System.Text;
+
+namespace Serilog.Sinks.File.Tests.Support
+{
+    /// <inheritdoc />
+    /// <summary>
+    /// Demonstrates the use of <seealso cref="T:Serilog.FileLifecycleHooks" />, by capturing the log file path
+    /// </summary>
+    class CaptureFilePathHook : FileLifecycleHooks
+    {
+        public string Path { get; private set; }
+
+        public override Stream OnFileOpened(string path, Stream _, Encoding __)
+        {
+            Path = path;
+            return base.OnFileOpened(path, _, __);
+        }
+    }
+}
--- a/src/Serilog.Sinks.File/Sinks/File/FileLifeCycleHookChain.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/FileLifeCycleHookChain.cs
@@ -29,10 +29,10 @@ namespace Serilog.Sinks.File
             _second = second ?? throw new ArgumentNullException(nameof(second));
         }
 
-        public override Stream OnFileOpened(Stream underlyingStream, Encoding encoding)
+        public override Stream OnFileOpened(string path, Stream underlyingStream, Encoding encoding)
         {
-            var firstStreamResult = _first.OnFileOpened(underlyingStream, encoding);
-            var secondStreamResult = _second.OnFileOpened(firstStreamResult, encoding);
+            var firstStreamResult = _first.OnFileOpened(path, underlyingStream, encoding);
+            var secondStreamResult = _second.OnFileOpened(path, firstStreamResult, encoding);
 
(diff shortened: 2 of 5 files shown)


nblumhardt:
Looks good to me; which scenario are we aiming to hit with this, in particular?

augustoproiete:
@nblumhardt @cocowalla I just did what I should have done from the start and [opened an issue to discuss and add more context](https://github.com/serilog/serilog-sinks-file/issues/191).

test/Serilog.Sinks.File.Tests/Support/CaptureFilePathHook.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.IO;
+using System.Text;
+
+namespace Serilog.Sinks.File.Tests.Support
+{
+    /// <inheritdoc />
+    /// <summary>
+    /// Demonstrates the use of <seealso cref="T:Serilog.FileLifecycleHooks" />, by compressing capturing the log file path
```
nblumhardt, line 8: Nit, extra word "compressing" here

nblumhardt (approved):

test/Serilog.Sinks.File.Tests/Support/CaptureFilePathHook.cs
augustoproiete: Good catch! I fixed + amended the commit

nblumhardt:
:shipit:

Falco20019:
Is there a roadmap on when this gets released? We are waiting for this. Our use case is having encryption applied on the file. When appending, we need to read the initialization vector from the header. Since neither the underlying stream nor the file name is offered through `WriteCountingStream`, we have no way in appending existing files right now.

nblumhardt:
Hi @Falco20019 - have you tried the `-dev` package, and does it work for you? Feedback would be great, thanks 👍

Falco20019:
@nblumhardt Yes, I tried it with that and that‘s when I saw the problems I documented in #199. So sadly this was only part of the solution since the underlying stream is not allowing write permission as it already opened the file with `FileShare.Read`.

PR #204, opened 2021-02-19
Support more options for rolling interval

tkrafael:


nblumhardt:
Thanks for your PR! Because of limited reviewer time available, changes to this sink need to come with a full proposal including analysis of corner cases and other impacts, with time taken to understand the existing codebase, before we could commit to working through a PR. If you're still interested in working on this, raising an issue with all of the detail would be welcome.

See also https://github.com/serilog/serilog-sinks-file/pull/182#issuecomment-708050271 for comments on problems with an earlier attempt at this.

PR #205, opened 2021-02-19
More granular rolling intervals

tkrafael:
This PR allows more granular intervals.
I'm using Serilog as TDR (Transaction details record) and CDR (call detail record) and while minute logs are ok, minute rolling intervals would create too much files.

This PR allows intervals in 5, 10, 15 and 30 minute intervals.


tkrafael:
Reopen PR with more rolling intervals

PR #206, opened 2021-03-04
Not crash when path is temporarily not available

kid101skater:
A simple change to handle if a path becomes un-available such as network or via usb. This was still passing unit tests and did not show a major perfomance loss:

with changes:
Elapsed: 6401 ms
Size: 58000000

original code:
Elapsed: 6199 ms
Size: 58000000

### DIFF
--- a/src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs
@@ -35,6 +35,8 @@ namespace Serilog.Sinks.File
         readonly bool _shared;
         readonly bool _rollOnFileSizeLimit;
         readonly FileLifecycleHooks _hooks;
+        private bool _pathPresent;
+        private string _directory;
 
         readonly object _syncRoot = new object();
         bool _isDisposed;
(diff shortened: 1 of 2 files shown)


nblumhardt:
Thanks for the PR!

I'm not super keen on this change, though - it's tempting to try to handle all of these cases, but I don't think the complexity is worth it, over the long run.

The perf impact here seems like it should/could be much higher than your measurements suggest, since we'd previously skip the whole file alignment process except at checkpoints, while with the change it seems like we'll check for path existence on every event.

kid101skater:
I think the perf here is a minimal change. This will only check if the root path “d:” for example comes offline. 

Could also add a setting to enable this if preferred. I think this is quite valuable as If the drive becomes unavailable then available again it will no longer capture logging.

nblumhardt:
Thanks for your reply; nothing's ever particularly minimal or straightforward with this sink - it runs in a massive number of environments and configurations 😅.

We've chosen to avoid too many special cases so that it's still possible to easily fork the sink and add whatever is needed - which I think would be a good solution for the scenario here.

nblumhardt:
Thanks again for the proposal!

PR #224, opened 2021-06-15
Add more rolling file intervals

tkrafael:
This PR will allow users to using following rolling intervals:
- Week: #182 
- Minute
- Five Minutes
- Ten Minutes
- Quarter Hour
- Half Hour
This will allow scenarios where this library can be used in scenarios other then logging (CDR generation, for example) and load those files when this period has passed.

Current intervals are either too sparse (hourly) or too granular (minutely) for this kind of scenario.


nblumhardt:
Hi! Thanks for the PR. When using `Week`, if I run my app today, shut it down, and then start it up again "tomorrow", won't I end up with a new file using these changes? (Think this was also the case for https://github.com/serilog/serilog-sinks-file/pull/182#issuecomment-708050271 ).

Issue #225, opened 2021-06-17
RollingInterval more extendable


blowin:
Replace enum with class, and replace extension with method. This will allow you to more flexibly customize the spacing and format. 

In my project, I need to use files with a specific name, but the library does not allow me to specify this. Therefore, I had to abandon this library and use a self-written solution.

I suggest the following solution:

 ```csharp
public interface IRollingInterval
{
    string Format { get; }

    DateTime? CurrentCheckpoint(DateTime instant);

    DateTime? NextCheckpoint(DateTime instant);
}
 ```

With implementation: InfiniteRollingInterval, YearRollingInterval etc

This solution allow easy extend behaviour. To avoid backward compatibility issues, you can make a separate interface overload and mark the enum as obsolete.

bartelink:
Realistically the current high level approach of defining rolling intervals is [not something that's viable to change given the broad use of this sink](https://github.com/serilog/serilog-sinks-file/issues/243#issuecomment-1404297429)

Closing as a duplicate of #243 and/or #298 - [the most viable way for anything to change is to have a fork demonstrate something that works, take the learnings from that, map it to a feature request here, and have someone implement it](https://github.com/serilog/serilog-sinks-file/issues/298#issuecomment-1762663305)

PR #226, opened 2021-06-20
#225 RollingInterval more extendable

blowin:


nblumhardt:
Thanks for your PR! Because of limited reviewer time available, substantial changes to this sink need to come with a full description including analysis of corner cases and other impacts, before we could commit to working through a PR. If you're still interested in working on this, raising a new issue with all of the detail would be welcome.

See also https://github.com/serilog/serilog-sinks-file/pull/182#issuecomment-708050271 and https://github.com/serilog/serilog-sinks-file/pull/224#issuecomment-865494836 for comments on problems with earlier attempts at this.

PR #241, opened 2021-09-20
Update Project URL so it links to repo

Sjoerdsjoerd:
I updated the Project URL so you can easily access the project from NuGet.
The same way as other Serilog packages, for example Serilog.Extensions.Logging:
afbeelding


nblumhardt:
LGTM - thanks 👍

PR #263, opened 2022-08-16
Added intervalMultiplier

foxjazz:
Added a multiplier to the interval option


nblumhardt:
Thanks for the suggestion. It's not likely that we'll add this feature at this point, though hopefully something you can apply and maintain via a downstream fork without too much trouble.

bartelink:
Thanks for taking the time to supply a PR demonstrating the idea - I appreciate the effort, and it's a long way from a drive by "can you add feature X already" comment.

However, I'm spring cleaning the issues and PRs list, and I feel that sadly the best thing to do here is to be cruel to be kind and close this for now.

See https://github.com/serilog/serilog-sinks-file/issues/298#issuecomment-1762663305 for my reasoning

TL;DR
- The high level approach of using an enum for the rolling interval is settled for now
- multipliers and/or other variations to naming is not likely to happen in the near term
- any resolution to the above should involve:
  1. writing up a detailed proposal as to how the facility would work
  2. ideally, a prototype, or links to a working fork that implements something very close to the proposal

PR #295, opened 2023-09-06
Add net6.0 target; update test TFMs

zachrybaker:
It's time to support .net 6, and .net 7. .net5.0 is EOL over a year ago.

### DIFF
--- a/src/Serilog.Sinks.File/Serilog.Sinks.File.csproj
+++ b/src/Serilog.Sinks.File/Serilog.Sinks.File.csproj
@@ -4,7 +4,7 @@
     <Description>Write Serilog events to text files in plain or JSON format.</Description>
     <VersionPrefix>5.0.1</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
-    <TargetFrameworks>net45;netstandard1.3;netstandard2.0;netstandard2.1;net6.0;net7.0</TargetFrameworks>
+    <TargetFrameworks>net45;netstandard1.3;netstandard2.0;netstandard2.1;net5.0;net6.0;net7.0</TargetFrameworks>
     <LangVersion>8.0</LangVersion>
     <Nullable>enable</Nullable>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
--- a/src/Serilog.Sinks.File/Serilog.Sinks.File.csproj
+++ b/src/Serilog.Sinks.File/Serilog.Sinks.File.csproj
@@ -4,7 +4,7 @@
     <Description>Write Serilog events to text files in plain or JSON format.</Description>
     <VersionPrefix>5.0.1</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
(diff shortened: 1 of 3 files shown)


bartelink:
Seealso my wall of text re the sister PR (some is related to this): https://github.com/serilog/serilog-sinks-console/pull/145#issuecomment-1709862581

src/Serilog.Sinks.File/Serilog.Sinks.File.csproj
```diff
@@ -4,7 +4,7 @@
     <Description>Write Serilog events to text files in plain or JSON format.</Description>
     <VersionPrefix>5.0.1</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
-    <TargetFrameworks>net45;netstandard1.3;netstandard2.0;netstandard2.1;net5.0</TargetFrameworks>
+    <TargetFrameworks>net45;netstandard1.3;netstandard2.0;netstandard2.1;net5.0;net6.0;net7.0</TargetFrameworks>
```
bartelink, line 7: See comment on PR with link re rationale for not (ever) adding `net7.0` here
```suggestion
    <TargetFrameworks>net45;netstandard1.3;netstandard2.0;netstandard2.1;net5.0;net6.0</TargetFrameworks>
```

test/Serilog.Sinks.File.Tests/Serilog.Sinks.File.Tests.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
 
-    <TargetFrameworks>net48;net5.0</TargetFrameworks>
+    <TargetFrameworks>net48;net6.0</TargetFrameworks>
```
bartelink, line 5: agree with this change
could add net7.0 though at this point its low value
would not object to adding net8.0 assuming CI has it

src/Serilog.Sinks.File/Serilog.Sinks.File.csproj
bartelink: NOTE this is on the assumption that we are aiming for a 5.0.1 or 5.1.0
If we were doing a 6.0 release of the sink:
- we support `netstandard2.0`, so no need to support `netstandard2.1`,`net5.0`,`net6.0`
- ... so I'd be proposing to simply remove `net5.0` and `netstandard2.1` here (f there is something that we support that benefits from either, then we add `net6.0` (and not `net8.0`))

bartelink (changes requested):
Looks good; just one change per my policy _proposal_ (will adjust review to match proposal as that goes).
On the assumption the aim is to get something out pretty quick, I've reviewed on the basis we are are doing a 5.0.1/5.1.0

zachrybaker (commented):
Thanks, sounds good.

zachrybaker (commented):
Looks good. Is there anything you need from me?

bartelink (approved):
LGTM now; can be merged
@zachrybaker once its merged, there will be a dev package automatically produced

General policy is that it gets tried and then subsequently merged to main for release, some low single or double digit days after it's validated

PR #297, opened 2023-10-02
Added pattern {date} and {seq} to file path in rolling sink.

GrzegorzRussek:
You now can use `{date}` and `{seq}` to define where date and sequence number will be stored. You can also alter date format by using `{date:yyyy-MM-dd}` for example.

### DIFF
--- a/src/Serilog.Sinks.File/Sinks/File/PathRoller.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/PathRoller.cs
@@ -16,6 +16,8 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using System.Linq;
+using System.Text;
 using System.Text.RegularExpressions;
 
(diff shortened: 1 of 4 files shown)


nblumhardt:
Thanks for the PR; we went down this path in the original Serilog.Sinks.RollingFile package, which is still maintained but is now deprecated. It would be interesting to see this approach developed further, but would need to be in a separate package, I don't think we want to take the current file sink in that direction.

bartelink:
Closing this on the same basis as https://github.com/serilog/serilog-sinks-file/pull/263#issuecomment-1770367986 in the interests of having the Issues and PR lists be short and clear

In short: Custom naming/interval multiplier policies are a non-trivial ask that should go through a prototype/fork + feature proposal gate before people invest time in PRs, as it's very frustrating for people implementing them to end up with something that realistically won't be merged.

PR #336, opened 2025-02-01
Make DefaultOutputTemplate public

### LINKED ISSUE #335: Suggestion: Make default template public
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. For example, "I'd like to do _x_ but currently I can't because _y_ [...]".

I'd like to keep the existing template almost entirely as-is, including across future Serilog updates. But, I'd like to modify one aspect of it, which I'm willing to take on myself.

However, in `FileConfigurationExtensions.cs`, `DefaultOutputTemplate` is a private constant rather than a public one, so I'm unable to programmatically modify it to suit my needs. I must instead copy/paste it.

https://github.com/serilog/serilog-sinks-file/blob/dev/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs#L35

**Describe the solution you'd like**

I'd like `DefaultOutputTemplate` to be a public constant. This way I can reference it in constructing my own template.

**Describe alternatives you've considered**

Copying/pasting `DefaultOutputTemplate` into my code and adding a comment indicating where I got it from.

**Additional context**

I want to output the HttpRequestId that comes from `.Enrich.WithHttpRequestId()` as part of this particular log file, to make tracing through files easier. So I'm looking to do something along the lines of:

```csharp
var template = FileLoggerConfigurationExtensions.DefaultOutputTemplate
        .Replace("[{Level:u3}] ", "[{Level:u3}] [Request {HttpRequestId}] ");

// Use template in Serilog
```

I'll submit a PR on this just in case you're interested, as the work itself is small. I couldn't find any historical precedent on whether it is considered acceptable to make the template public.

SeanKilleen:
Resolves #335

### DIFF
--- a/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
+++ b/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
@@ -32,7 +32,7 @@ public static class FileLoggerConfigurationExtensions
 {
     const int DefaultRetainedFileCountLimit = 31; // A long month of logs
     const long DefaultFileSizeLimitBytes = 1L * 1024 * 1024 * 1024; // 1GB
-    const string DefaultOutputTemplate = "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{Level:u3}] {Message:lj}{NewLine}{Exception}";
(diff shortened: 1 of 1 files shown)


SeanKilleen:
There are errors on the AppVeyor build here but as far as I can tell they aren't related to my changes.

nblumhardt:
Closing following the discussion in #335. Thanks again!

PR #342, opened 2025-03-08
Support `ILoggingFailureListener`

nblumhardt:
Seals `PeriodicFlushToDiskSink`, so technically a breaking change.

Includes a switch to Actions for CI.


src/Serilog.Sinks.File/Sinks/File/RollingFileSink.cs
```diff
@@ -84,12 +87,18 @@ public void Emit(LogEvent logEvent)
                 AlignCurrentFileTo(now, nextSequence: true);
             }
 
-            /* TODO: We REALLY should add this to avoid stuff become missing undetected.
-            if (_currentFile == null)
-            {
-                SelfLog.WriteLine("Log event {0} was lost since it was not possible to open the file or create a new one.", logEvent.RenderMessage());
-            }
-            */
+            failed = _currentFile == null;
```
nblumhardt, line 90: @Falco20019 this should cover the TODO

Falco20019: I will have a look beginning next week, thanks 🎉

Falco20019: @nblumhardt There is sadly no artifact to use on AppVeyor as the build "failed successfully" ;)
https://ci.appveyor.com/project/serilog/serilog-sinks-file/builds/51655364

I will create a local build for testing for now, but you might want to fix the CI pipeline.

src/Serilog.Sinks.File/Sinks/File/FailedSink.cs
Falco20019, line 1: I do like the concept of having every log still appearing somewhere if everything went wrong. Thanks!

Falco20019 (approved):
Did some testing and seems to work as expected. Thanks!

Could you still have a hotfix-release before having this major release? Our company policy requires some effort on switching major version, so for getting the bug fixed first on our current version would help me a lot!

Build.ps1
```diff
@@ -1,44 +1,79 @@
...
-$revision = @{ $true = "{0:00000}" -f [convert]::ToInt32("0" + $env:APPVEYOR_BUILD_NUMBER, 10); $false = "local" }[$NULL -ne $env:APPVEYOR_BUILD_NUMBER];
-$suffix = @{ $true = ""; $false = "$($branch.Substring(0, [math]::Min(10,$branch.Length)))-$revision"}[$branch -eq "main" -and $revision -ne "local"]
+    Write-Output "build: Package version prefix is $versionPrefix"
 
-Write-Output "build: Package version suffix is $suffix"
+    $branch = @{ $true = $env:CI_TARGET_BRANCH; $false = $(git symbolic-ref --short -q HEAD) }[$NULL -ne $env:CI_TARGET_BRANCH];
+    $revision = @{ $true = "{0:00000}" -f [convert]::ToInt32("0" + $env:CI_BUILD_NUMBER, 10); $false = "local" }[$NULL -ne $env:CI_BUILD_NUMBER];
+    $suffix = @{ $true = ""; $false = "$($branch.Substring(0, [math]::Min(10,$branch.Length)) -replace '([^a-zA-Z0-9\-]*)', '')-$revision"}[$branch -eq "main" -and $revision -ne "local"]
+    $commitHash = $(git rev-parse --short HEAD)
+    $buildSuffix = @{ $true = "$($suffix)-$($commitHash)"; $false = "$($branch)-$($commitHash)" }[$suffix -ne ""]
```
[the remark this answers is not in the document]
Falco20019: Personally I would use `7.0.0-ci.1+04ac85d-selflog-on-open-failure`, but tried to keep the suggestion closer to what you do right now.

src/Serilog.Sinks.File/Sinks/File/SharedFileSink.OSMutex.cs
```diff
@@ -149,13 +165,12 @@ bool TryAcquireMutex()
...
             if (!_mutex.WaitOne(MutexWaitTimeout))
             {
-                SelfLog.WriteLine("Shared file mutex could not be acquired within {0} ms", MutexWaitTimeout);
                 return false;
             }
         }
         catch (AbandonedMutexException)
         {
-            SelfLog.WriteLine("Inherited shared file mutex after abandonment by another process");
+            SelfLog.WriteLine("inherited the shared file mutex after abandonment by another process");
```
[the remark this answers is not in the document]
nblumhardt: Thanks for taking a look. Yes, this one doesn't go through the failure listener because it's not a failure. In this situation, another process failed, but we've picked up the mutex and we're carrying on.

Build.ps1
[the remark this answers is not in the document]
nblumhardt: Thanks for digging into this! These scripts are from the upstream `serilog/serilog` and we're trying to avoid configuration drift by making any changes up there. I agree this could be better 👍  - I'll leave this as-is here but take a look in the next round of maintenance on the Serilog package.

nblumhardt:
@Falco20019 thanks for taking a look a this. I am not sure about the timeline for releasing a patch, it's not hugely likely we'd push a point release out, as some additional work would be needed to get the original AppVeyor build through, and each new version causes some additional effort for downstream consumers.

Since `FileSink` and `SharedFileSink` are public, one option might be to continue with 6.0.0 and embed a copy of `RollingFileSink.cs` plus a couple of the minimal supporting types (`PathRoller` comes to mind) in your project, until 7.0.0 is published?

Falco20019:
It's fine, I just asked in case you are fine with it. As we are using an internal NuGet Repository, I will just publish a hotfix version there to be used and switch back to official releases in November. Not the first time we have to do it and surely not the last.

Issue #243, opened 2021-09-25
Customize filename while using rolling policies


Deilan:
Requesting a capability to customize a filename while using [rolling policies](https://github.com/serilog/serilog-sinks-file#rolling-policies).

zivfuture:
any update?

osnipezzini:
Any News?

bartelink:
An 'update' and/or 'news' would require a) an agreed spec b) a PR implementing that

- The first bit can be helped by people providing details of their requirements and/or a proposed API - this is a reasonable place for such a discussion to take place (given this sink has very broad usage, so it's worth getting the design right such that it covers the needs of as many people as possible; removing misfeatures and/or changing behavior afterward is not going to be viable).
- The second bit is a matter of someone somewhere making the time to implement it.

ReikoTigon:
So it's not possible to let the rolling policy save the file in another folder than the original one? Because I really would need that to keep it organized.

cocowalla:
@ReikoTigon maybe this plugin would work for you: [cocowalla/serilog-sinks-file-archive](https://github.com/cocowalla/serilog-sinks-file-archive) ?

ReikoTigon:
Hmm if i get the plugin correctly it moves them instead of deleting them?
What I want is use the defualt mechanism by srilog to apply the daily rolling pollicy with 7 days, but tho move those old 7 logs to a dedicated "logs" folder, whereas the current log is located in the main folder.

or be able to name the current log sth like "current.log" and all the old logs "log_DATE.log" like it would do atm.

wzuqui:
in NLog there are some properties like achiveEvery and achiveDateFormat that only modifies archived files, can it be possible to customize the same for serilog-skins-file?

see: https://nlog-project.org/documentation/v3.2.1/html/Properties_T_NLog_Targets_FileTarget.htm

wzuqui:
In NLog, there are some properties such as "archiveEvery" and "archiveDateFormat" that only modify archived files. Is it possible to customize the same for serilog-sinks-file?

see: https://nlog-project.org/documentation/v3.2.1/html/Properties_T_NLog_Targets_FileTarget.htm

nblumhardt:
Hi folks!

[...]

PR #343, opened 2025-03-10
Added support for customising date in filename

RandomTIP:
Address (Partially): #243 

Changes: 

- PathRoller now recognizes and uses {Date} placeholder in file name.
- Regex matching and directory search patterns have been adjusted for use with {Date} placeholder

This update should Enable users to use following configuration: 

```
var config = new LoggerConfiguration()
            .WriteTo.File("{Date}-log.txt", rollingInterval: RollingInterval.Day);
var log = config.CreateLogger();
```

Unit Tests Added and Updated in:

- TemplatePathRollerTests
- RollingFileSinkTests

### DIFF
--- a/src/Serilog.Sinks.File/Sinks/File/PathRoller.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/PathRoller.cs
@@ -43,16 +43,9 @@ sealed class PathRoller
         _directory = Path.GetFullPath(pathDirectory);
         _filenamePrefix = Path.GetFileNameWithoutExtension(path);
         _filenameSuffix = Path.GetExtension(path);
-        _filenameMatcher = new Regex(
-            "^" +
-            Regex.Escape(_filenamePrefix) +
-            "(?<" + PeriodMatchGroup + ">\\d{" + _periodFormat.Length + "})" +
-            "(?<" + SequenceNumberMatchGroup + ">_[0-9]{3,}){0,1}" +
-            Regex.Escape(_filenameSuffix) +
-            "$",
-            RegexOptions.Compiled);
+        _filenameMatcher = GetFileNameMatcher();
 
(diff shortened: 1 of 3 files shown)


RandomTIP:
I know this PR came out of nowhere, but I needed support for this for my project and thought I would send a PR to the original repo as well since the changes are quite simple and straightforward. 

This partially addresses #243 without any sort of high level customization, I wanted to make it as simple as possible because, after one or two days worth of searching, I think most of the requests were only related to Date customization in filename. 

I would appreciate and take into considaration any and all notes on this one

nblumhardt:
Thanks @RandomTIP. This functionality is something we specifically decided to leave behind in Serilog.Sinks.RollingFile - using that package is the way to go, if you need `{Date}`, but otherwise, using your own fork of this sink is going to be the way forwards. Thanks again!

RandomTIP:
Thanks for the information. I was not able to gather that much myself for some reason... I assumed Serilog.Sinks.RollingFile was deprecated and that is why I overlooked it.

nblumhardt:
It is deprecated, but still works; grabbing the `RollingFileSink` class from there, but modifying it to use `Serilog.Sinks.File`'s `FileSink`/`SharedFileSink` should cover a lot of the implementation details and result in a fully-maintained option. HTH!

BernhardMarconato:
@nblumhardt I need a very similar feature. I tried copying the `Serilog.Sinks.File.RollingFileSink` to my project to create a fork, but a lot of the members used are `internal`, making it hard to have a 1:1 copy just with a different `PathRoller`. 

Wouldn't it be possible to make `PathRoller` implement a public `IPathRoller` interface, and let users provide a custom `IPathRoller` implementation? I don't think this would cause much additional maintenance effort from your side.

Issue #40, opened 2017-11-14
Fixed filename with rolling archive files


bartkusters:
When logging rolls, all new entries are written to the rolled file. Can't this be changed so that logging keeps getting written to "sample.log" and the rolled files "sample001.log", "sample002.log", etc. keep the history?

hali4831:
This is something I am looking for too

jawadatgithub:
Me too, to explain the benefit of that, I usually open the log.txt in notepad++ to keep monitoring current logs, and in case of issue happened while ago, I go to rolled file based on last modified date.

So most of the times I just need to have log.txt opened, and rarely check rolled files. and that's the default behavior used in nlog, but for .net core 2 apps I still prefer serilog.

bitchkat:
I was actually just going to ask about this exact same behavior.  The way that bartkusters describes is very traditional -- at least in the linux/unix world.  If I'm tailing a log file, it just silently stops when the log file rolls over.  Also you have to ls all the files to find the last one to start a tail when the log file has been rolled over.

nblumhardt:
Thanks for all the input :+1: 

Is anyone interested in exploring this one a bit further, to see what the implementation effort would involve, and how the file logger would need to change in order to incorporate this scenario?

nblumhardt:
See also some earlier discussion in: https://github.com/serilog/serilog-sinks-rollingfile/issues/26

bartelink:
While its true that log4net and NLog both do renaming stuff (at least by default), I'd like to point out the following reasons to maintain [at a minimum, backcompat with] the  status quo:

- I've seen cases where that's problematic (e.g. backup, viewing or AV software can block a rolling set of renames)
- doing such a rename in the context of concurrent writing to a shared resource is fraught with danger
- log forwarders in general (and def Splunk one) needs to go to effort to fingerprint content to keep track of it all
- Splunk and/or other files can not track the renamed state of the file whereas this scheme keeps it all predictable - if I want to track back to a source file it's there as long as it has not been aged out.

But... the reason I'm here is to ask for a feature/tweak... @nblumhardt or others, is anyone interested in having rolling by size trigger usage of suffix universally? i.e. if the name requested is `x/app.log` then the first file written will be x/app_000.log`. This would make it easier to explain a naming pattern to someone ingesting and/or avoid surprises vs having to explain the special casing. While not trivial, it does not look crazy difficult to impl based on the current code.

nblumhardt:
[...]

PR #345, opened 2025-03-17
Add ability for static base file and customizable file roll patterns.

jublin:
In my work, a number of our projects use Serilog. One project has a task to revamp the logging to use items like Elastic/FluentD, but must also still output to a static file for developer/support engineer use in the field. 

Given that a number of projects were using Serilog, this seemed an easier feature to add than to figure out other logging libraries/roll our own. Other sinks forked off of this removed the NETStandard 2.0 compatibility, which was a no-go for use on our end. 

Did this in my own time both to help my work, but to have something I can use outside of work on my tinkering projects. 

Adds 3 optional parameters: 
 - **keepPathStaticOnRoll** (`bool`) - Keep the base file name passed in (`log.txt` for instance) as the base file. 
 - **customFormatFunc** (`Func<DateTime?,string>`) - Function to give a custom format for rolling files. The yyyyMMDD format did not work for requirements. 
 - **customRollPattern** (`string`) - A string regex pattern that will match the output of customFormatFunc. This must be provided when custom format is used, as the rollovers will not work (nor will it validate).

### DIFF
--- a/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
+++ b/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
@@ -155,7 +155,8 @@ public static class FileLoggerConfigurationExtensions
         Encoding encoding)
     {
         return File(sinkConfiguration, path, restrictedToMinimumLevel, outputTemplate, formatProvider, fileSizeLimitBytes, levelSwitch, buffered,
-            shared, flushToDiskInterval, rollingInterval, rollOnFileSizeLimit, retainedFileCountLimit, encoding, null);
+            shared, flushToDiskInterval, rollingInterval, rollOnFileSizeLimit, retainedFileCountLimit, encoding, null,
+             null);
     }
 
     /// <summary>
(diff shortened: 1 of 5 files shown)


nblumhardt:
Thanks for jumping in! This feature [has been discussed here in the past](https://github.com/serilog/serilog-sinks-file/issues/40) and, while it's a useful feature, it comes with some complexity and implementation challenges we're not keen to take on in this sink.

A fork supporting this feature has been developed in https://github.com/dfacto-lab/serilog-sinks-file, and I think contributions/improvements over there would be welcomed (best to ping the maintainer and check what's currently on the list :-))

nblumhardt:
Thanks for the follow up 👍 

Keeping things healthy/maintained over a long time span is definitely an unsolved problem, unfortunately. Being really strategic about feature additions is how this one's still going.

Forking the fork (or this project) might be the easiest option at this point. A maintained wrapper around this project that reuses its `FileSink` but adds new rolling policies would be really appreciated out there and might be easier to update long-term, in case that's an option.

Issue #247, opened 2022-01-20
Explicit Flushing 


JLorch:
We have a scenario where we want to use a buffered FileSink with a flush interval of lets say 5 seconds. If a critical error (above a certain log level) is logged we want to increase our chances of logging this to file (as the application might be unstable at this point and the periodic flush might not happen) and performing an explicit flush of the log. 

Is this possible to achieve?

nblumhardt:
Thanks for the note; no, there's no explicit flushing feature implemented today. Are the errors you're interested in `Fatal`/`Critical`? I don't think we'd expose an API for this, but it could be interesting to investigate flushing automatically when an event at this level is observed.

JLorch:
Yes, the error levels that would make sense for our use case is probably Error and Fatal. Information that could be very valuable for fault tracing.

nblumhardt:
Thanks for your reply. I think `Error` could be problematic - generating a large number of errors is not uncommon, and adding flush overhead to these could make a bad situation worse. `Fatal` is different in that it is supposed to communicate a probable-termination or similar scenario, where a flush to disk should usually make sense.

Thinking some more, I wonder if there's a way to implement this today using a wrapper sink and `FileLifecycleHooks`?

The wrapper sink would wrap the file sink, and pass in its lifecycle hooks implementation, which would give it a chance to capture a reference to the underlying file stream. Then, after forwarding an event through to the wrapped sink, the wrapper could flush the file, if the level of the event was error/fatal? 🤔

nblumhardt:
Closing as I think the workaround above is the best current option; if you're still keen to implement this, let us know, happy to help work through it.

BernhardMarconato:
Using `FileLifecycleHooks` is no suitable workaround, as you can only access the `Stream` and flush it, but for example the `FileSink` also needs to flush the `TextWriter`. 

I would propose to add something alike:

```c#
public static LoggerConfiguration File(this LoggerSinkConfiguration sinkConfiguration, LogEventLevel forceFlushAtMinimumLevel, ...)
```

The default value would be `Fatal`, but users could disable it completely or even force to flush all the time.
@nblumhardt What are your thoughts? I am willing to create a PR for it. Thanks!

PR #356, opened 2025-11-02
Always flush sink on Fatal events

BernhardMarconato:
Implements https://github.com/serilog/serilog-sinks-file/issues/247

Always flush the `FileSink` when a `LogLevelEvent.Fatal` event is emitted. This is a breaking change in behavior.

A common use case is having a flush interval set, but still wanting to force flush `Fatal` events to the file before the process terminates. Previously, these log events would have been lost.

~~Add the option to provide a minimum log level using `flushAtMinimumLevel` for which the sink should flush to disk immediately.
To retain backwards compatibility, the default value is `LevelAlias.Off`.~~


nblumhardt:
Hi @BernhardMarconato!

Thinking about the scenario you describe, it sounds like that one in particular would be solved by calling `FlushToDisk()` in `FileSink.Dispose()`: when a fatal error is logged, generally it's within some kind of unwinding which can also close/flush the logging pipeline. Flushing to disk in there should have the same effect, without requiring any additional API. What do you think?

BernhardMarconato:
Thanks for taking a look!  

You’re right that the filestream should also be flushed during disposal. However, in most cases, flushing the TextWriter (as is currently done) seems to be sufficient and flushes to the file as well. That’s what I’m currently doing by calling `Serilog.Log.CloseAndFlush();`. But due to platform limitations (WinUI), I can’t guarantee that the app will actually terminate upon receiving an unhandled exception. If the logger is disposed eagerly, it might stop logging too soon.

I’m not sure how you feel about breaking changes, but I think it could be justified to always flush for Fatal events, even without an additional API for configuring this.

Alternatively, I’d be happy if there was a way to access the `IFlushableFileSink` instance directly so I could flush manually when needed. As far as I know, this isn’t possible at the moment due to the nested sink structure (but please correct me if I’m wrong).

nblumhardt:
Thanks for the analysis and suggestions, much appreciated.

Having fatal errors fully flushed to disk, even when buffering is disabled, seems like it would be acceptable and a decent improvement, given that there's no plan to expose flushing via another mechanism. E.g. thinking it would be:

```csharp
if (logEvent.Level == LogEventLevel.Fatal)
    FlushToDisk();
else if (!_buffered)
    _output.Flush();
```

It does open up the possibility of causing a performance drain if a large number of `Fatal` errors are logged in quick succession, but at that point the wheels are likely to be falling off anyway. Interested in opinions on this if anyone following along thinks otherwise.

It'd be worth bumping the major version and documenting this in the `WriteTo.File()` XDOC.

BernhardMarconato:
Thanks @nblumhardt, I applied your suggestions! Let me know if anything else is required.

nblumhardt:
Thanks! Apologies for the slow turnaround. Looks good 👍

BernhardMarconato:
Hi @nblumhardt, I wanted to ask when this change could be released? I saw in other repos that you upgraded to .NET 10 before a release, so I created a PR for this in this repo, in case it helps. https://github.com/serilog/serilog-sinks-file/pull/359 Thanks!

PR #360, opened 2026-02-18
Change constants to public in FileLoggerConfigurationExtensions

InteXX:
Proposal: scope the default values to public, so that consuming applications may more easily subclass `MessageTemplateTextFormatter` as a side-by-side for `JsonFormatter`.

Example usage under the current construct (brittle):
```
using System.Reflection;
using Serilog;
using Serilog.Formatting.Display;

namespace TestProject.Intexx.Extensions.Logging.Text
{
  public class TextFormatter : MessageTemplateTextFormatter
  {
    public TextFormatter() : base(DefaultTemplate) {}

    private static string DefaultTemplate
    {
      get
      {
        return typeof(FileLoggerConfigurationExtensions).Assembly
          .GetType($"{nameof(Serilog)}.{nameof(FileLoggerConfigurationExtensions)}")
          .GetField("DefaultOutputTemplate", BindingFlags.NonPublic | BindingFlags.Static)
          .GetValue(null);
      }
    }
  }
}
```

### DIFF
--- a/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
+++ b/src/Serilog.Sinks.File/FileLoggerConfigurationExtensions.cs
@@ -30,9 +30,9 @@ namespace Serilog;
 /// <summary>Extends <see cref="LoggerConfiguration"/> with methods to add file sinks.</summary>
 public static class FileLoggerConfigurationExtensions
 {
-    const int DefaultRetainedFileCountLimit = 31; // A long month of logs
-    const long DefaultFileSizeLimitBytes = 1L * 1024 * 1024 * 1024; // 1GB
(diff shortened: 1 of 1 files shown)


nblumhardt:
Thanks for the suggestion! Sorry about the slow reply, busy week :-)

In the context they're used, they're not particularly special values, just defaults that are used when the consumer doesn't really have any specific requirements. I think it's a reasonable proposal, but up against the goal of keeping public surface area to a minimum I don't think they quite get there. Embedding the string ` "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{Level:u3}] {Message:lj}{NewLine}{Exception}"` in your scenario is more robust and stable than the reflection-based approach, and would arguably provide some insulation from unwanted changes in Serilog's defaults.

If you're building a custom text formatter of any size, it might be worth checking out https://nblumhardt.com/2021/06/customize-serilog-text-output/ - the infrastructure in _Serilog.Expressions_ is much better than what comes out of the box in the base _Serilog_ package. HTH!

InteXX:
Got it, thanks.

PR #368, opened 2026-05-12
Silence ObjectDisposedException race in PeriodicFlushToDiskSink

### LINKED ISSUE #367: PeriodicFlushToDiskSink races RollingFileSink on roll, spams failure listener
## Repro

Application configured as:

```csharp
WriteTo.File(
    "logs/app-.log",
    rollingInterval: RollingInterval.Day,
    flushToDiskInterval: TimeSpan.FromSeconds(1))
```

Under steady traffic. On each daily roll (or size-based roll when configured), the periodic flush timer in `PeriodicFlushToDiskSink` fires while `RollingFileSink` is between closing the old `FileSink` and opening the new one, producing:

```
Serilog.Sinks.File.PeriodicFlushToDiskSink: could not flush the underlying file to disk (Temporary, unspecified events)
System.ObjectDisposedException: Cannot access a closed file.
   at Serilog.Sinks.File.RollingFileSink.FlushToDisk()
   at Serilog.Sinks.File.PeriodicFlushToDiskSink.FlushToDisk(IFlushableFileSink flushable)
```

This is a `Temporary` failure from the listener's point of view (the sink does recover on the next interval) but the noise volume is significant whenever `SelfLog` (the default listener) is persisted, surfaced to operators, or fed into another telemetry stream.

Concrete numbers from a downstream telemetry pipeline that tees `SelfLog` into a diagnostic file: 22 155 of 88 786 lines (~25 %) of the operator-visible diagnostics file were this single exception. A second deployment of the same software saw closer to 50 %. Real application-level events were drowned out.

## Why it is not a real failure

The current code already documents the race inline:

```csharp
// May throw ObjectDisposedException, since we're not trying to synchronize
// anything here in the wrapper.
flushable.FlushToDisk();
```

`ObjectDisposedException` at this site can only come from the underlying `FileSink` having been disposed by `RollingFileSink` while it swapped to a fresh sink. The next timer interval flushes the new sink successfully. There is no data loss and no operator action.

## Proposal

Catch `ObjectDisposedException` separately in `PeriodicFlushToDiskSink.FlushToDisk` and suppress the failure report. Real I/O failures (`IOException`, `UnauthorizedAccessException`, disk-full, etc.) keep flowing through the existing `Temporary` failure path.

Patch is ~5 LoC plus tests. PR forthcoming.

GioAbq:
Closes #367.

## Background

`PeriodicFlushToDiskSink.FlushToDisk` runs on a `Timer` independent of `RollingFileSink`'s roll lifecycle. When a roll closes the active `FileSink` and opens a new one, the timer can fire in between and catch the disposed instance. Today this is caught by the generic `catch (Exception)` branch and reported as a `Temporary` failure through `ILoggingFailureListener` (default: `SelfLog`), once per `flushToDiskInterval` per roll, indefinitely.

The race itself is already documented inline in the source:

```csharp
// May throw ObjectDisposedException, since we're not trying to synchronize
// anything here in the wrapper.
flushable.FlushToDisk();
```

## Change

Add a dedicated `catch (ObjectDisposedException)` clause before the generic one and suppress. Real I/O failures (`IOException`, `UnauthorizedAccessException`, etc.) still flow through the existing `Temporary` failure path.

`FlushToDisk(IFlushableFileSink)` is promoted from implicit `private` to `internal` so the new tests in `Serilog.Sinks.File.Tests` (already wired through `InternalsVisibleTo` in `AssemblyInfo.cs`) can drive it deterministically without relying on `Timer` wall-clock cadence. No public API surface changes.

## Why this is safe

- `ObjectDisposedException` at this site can only originate from the underlying `FlushToDisk()` call on a disposed `FileSink`. There is no other shared state on the path.
- The next timer tick picks up the new `FileSink` and flushes cleanly. No data loss.
- Behaviour is unchanged for any other exception type.

## Tests

Three new tests in `PeriodicFlushToDiskSinkTests`:

- `FlushToDisk_WhenFlushableThrowsObjectDisposed_SuppressesFailureReport` - listener receives no `OnLoggingFailed` call when the underlying flush throws `ObjectDisposedException`.
- `FlushToDisk_WhenFlushableThrowsIOException_ReportsAsTemporaryFailure` - listener still receives the exception as `Temporary` failure when the underlying flush throws `IOException`.
- `FlushToDisk_WhenNoEventEmitted_DoesNotInvokeUnderlyingFlush` - guards the existing `_flushRequired` short-circuit.

Tests pass on `net48`, `net8.0`, `net9.0` (3 / 3 each).

## Downstream impact

Real-world case from a downstream telemetry pipeline that tees `SelfLog` into a diagnostic file: 22 155 / 88 786 lines (~25 %) of the operator-visible diagnostics file were this single exception. A second deployment saw closer to 50 %. A local downstream filter mitigates the noise for that pipeline, but fixing upstream removes it for every consumer of `Serilog.Sinks.File` whose listener persists or surfaces `Temporary` failures.

### DIFF
--- a/src/Serilog.Sinks.File/Sinks/File/PeriodicFlushToDiskSink.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/PeriodicFlushToDiskSink.cs
@@ -89,10 +89,7 @@ public sealed class PeriodicFlushToDiskSink : ILogEventSink, IDisposable, ISetLo
         }
         catch (ObjectDisposedException)
         {
-            // Expected race against rolling/shutdown: the underlying sink was disposed between
-            // the timer firing and the flush call. Not an I/O failure; suppress to avoid spamming
-            // the failure listener (default: SelfLog). The next timer tick will flush the
-            // replacement sink cleanly.
+            // Avoid spamming failure listeners when rolling to a new file.
         }
         catch (Exception ex)
         {
--- a/src/Serilog.Sinks.File/Sinks/File/PeriodicFlushToDiskSink.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/PeriodicFlushToDiskSink.cs
@@ -76,7 +76,7 @@ public sealed class PeriodicFlushToDiskSink : ILogEventSink, IDisposable, ISetLo
         (_sink as IDisposable)?.Dispose();
     }
 
-    void FlushToDisk(IFlushableFileSink flushable)
+    internal void FlushToDisk(IFlushableFileSink flushable)
     {
         try
         {
@@ -86,6 +86,13 @@ public sealed class PeriodicFlushToDiskSink : ILogEventSink, IDisposable, ISetLo
                 // anything here in the wrapper.
                 flushable.FlushToDisk();
             }
+        }
+        catch (ObjectDisposedException)
+        {
+            // Expected race against rolling/shutdown: the underlying sink was disposed between
+            // the timer firing and the flush call. Not an I/O failure; suppress to avoid spamming
+            // the failure listener (default: SelfLog). The next timer tick will flush the
+            // replacement sink cleanly.
         }
         catch (Exception ex)
         {
--- a/test/Serilog.Sinks.File.Tests/PeriodicFlushToDiskSinkTests.cs
+++ b/test/Serilog.Sinks.File.Tests/PeriodicFlushToDiskSinkTests.cs
@@ -1,0 +1,93 @@
+using Serilog.Core;
+using Serilog.Events;
+using Serilog.Sinks.File.Tests.Support;
+using Xunit;
+
+#pragma warning disable 618 // PeriodicFlushToDiskSink is marked obsolete; we still test it directly.
+
+namespace Serilog.Sinks.File.Tests;
+
+public class PeriodicFlushToDiskSinkTests
+{
+    [Fact]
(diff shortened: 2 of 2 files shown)


src/Serilog.Sinks.File/Sinks/File/PeriodicFlushToDiskSink.cs
```diff
@@ -87,6 +87,13 @@ void FlushToDisk(IFlushableFileSink flushable)
                 flushable.FlushToDisk();
             }
         }
+        catch (ObjectDisposedException)
+        {
+            // Expected race against rolling/shutdown: the underlying sink was disposed between
+            // the timer firing and the flush call. Not an I/O failure; suppress to avoid spamming
+            // the failure listener (default: SelfLog). The next timer tick will flush the
+            // replacement sink cleanly.
+        }
```
nblumhardt, line 93: The comment here is a little wordy, could just be:

```
// Avoid spamming failure listeners when rolling to a new file.
```

nblumhardt (commented):
Looks good, thanks. Just one comment.

src/Serilog.Sinks.File/Sinks/File/PeriodicFlushToDiskSink.cs
GioAbq: Done, thanks!

GioAbq:
@nblumhardt Can we merge it?

nblumhardt:
Done; thanks @GioAbq :+1:

